PR #43, opened 2016-09-11
ApplyPass -> ApplyPasses; Refactored infer pass;

jermainewang:
Some refactors.

Question for @tqchen :

I have codes like following:
```python
net = mx.symbol.Variable('data')
net = mx.symbol.FullyConnected(net, name='fc1', num_hidden=40)

executor = net.simple_bind(mx.cpu(0), 'write', data=some_data_shape)
# ... copy data to the "data" placeholder
outputs = executor.forward()
executor.backward(outputs[0])
```
This is a symbol without a loss function, so I just put the output of the symbol as the output grad. When I run the InferShapePass, the output grad has NO shape. How can I easily fix that?


tqchen:
please fix https://travis-ci.org/dmlc/nnvm/jobs/159126401

PR #45, opened 2016-09-12
Fix inplace. Only permit shared when shape and dtype is the same

piiswrong:


tqchen:
shape check was too strict and this prevents the reshape optimization.

piiswrong:
No. It only checks for Size().

PR #46, opened 2016-09-14
[API] Change attr to explicit name set_attr

tqchen:


tqchen:
@piiswrong @jermainewang

PR #53, opened 2016-09-22
Fix infertype pass

piiswrong:


tqchen:
I think we might want to add logic to do a second pass to fix the types to 0, otherwise for codes that do not yet have infer type, the default type is not used

tqchen:
Please confirm mxnet/nnvm still work before merge this in

piiswrong:
@tqchen updated

tqchen:
Maybe a better way is to provide a default inference function,  in the case of type inference it sets the Attrs to be default. 

Note that the num_unknown need to be correct and current code is not. (for type inference, num_unknown remains

tqchen:
insert after https://github.com/dmlc/nnvm/blob/master/src/pass/infer_shape_type.cc#L112

For shape inference, if shape function is not presented, an error need to be raised

piiswrong:
@tqchen I'm a little confused,  do you have time to fix it?

tqchen:
https://github.com/dmlc/nnvm/pull/54

PR #57, opened 2016-09-23
[OP] Enable register via match tag

tqchen:
(edit) new api

```c++

NNVM_REGISTER_OP(add)
.include("ElementwiseOpAttr");

// register same attributes to all the ops that matchs the attribute condition.
NNVM_REGISTER_OP_GROUP(ElementwiseOpAttr)
.set_attr<FInferShape>("FInferShape", ElementwiseInferShape);

// mul will also have the ElementwiseInferShape 
NNVM_REGISTER_OP(mul)
.include("ElementwiseOpAttr");
```


tqchen:
@jermainewang @piiswrong

tqchen:
please see if the name  ```NNVM_REGISTER_BY_MATCH_ATTR``` makes sense to you, I think it is an OK name and cannot come up with a shorter version for now

piiswrong:
```
NNVM_REGISTER_OP(add)
.set_attr<std::string>("OpType", "Elementwise");

// register same attributes to all the ops that matchs the attribute condition.
NNVM_REGISTER_OP_GROUP(OpType, "Elementwise")
.set_attr<FInferShape>("FInferShape", ElementwiseInferShape);

// mul will also have the ElementwiseInferShape tag
NNVM_REGISTER_OP(mul)
.set_attr<std::string>("OpGroup", "Elementwise");
```
looks better

piiswrong:
BTW, what's the relation between OpType and TagType? It's confusing

piiswrong:
This makes it even harder to figure out where things are registered.

It would be MUCH easier to document and understand if you use a different interface for setting tags:
```
NNVM_REGISTER_OP(mul)
.add_to_group("Elementwise");
```
also instead of using names like Elementwise, it should be something unique so that a search in the code base will point to the registration point

piiswrong:
Actually I think
```
NNVM_REGISTER_OP(mul)
.include("ElemwiseOpAttrs");
```
Makes more sense

tqchen:
The idea is that we don't want to specialize the field group or tag. But simply use attribute. There can be multiple attributes that corresponding to a tag.

For example
```
NNVM_REGISTER_OP(add)
.set_attr<std::string>("OpType", "Elementwise")
.set_attr<std::string>("FuncType", "BinaryOp");

// register same attributes to all the ops that matchs the attribute condition.
NNVM_OP_GROUP_BY_MATCH_ATTR(OpType, "Elementwise")
.set_attr<FInferShape>("FInferShape", ElementwiseInferShape);

NNVM_OP_GROUP_BY_MATCH_ATTR(FuncType, "UnaryOp")
.set_attr<FXX>("FXX", UnaryFXX);

// mul will also have the ElementwiseInferShape tag
NNVM_REGISTER_OP(mul)
.set_attr<std::string>("OpType", "Elementwise");

NNVM_REGISTER_OP(exp)
.set_attr<std::string>("OpType", "Elementwise")
.set_attr<std::string>("FuncType", "Unary");


```

piiswrong:
That makes it more complicated but doesn't add much utility. `include(OpType:Elementwise)` would to the same thing.

I think it definitely need a different api. you can still implement it via attrs internally.

piiswrong:
Consider doing it this way: an op can include another op and copy all it's attrs. This can be done recursively and groups can also be registered as ops

tqchen:
Group is definitely different from what op is(they have incomplete information, and partial info about something), so it does not make sense to do so.

```NNVM_OP_GROUP_BY_MATCH_ATTR``` provides a semantics like

```c++
NNVM_FOR_EACH_OP_WHOSE_ATTR_EQ(AttrName, Attr_Value)
DO
op.set_attr<bla>()
END
```
Which is quite clear,

piiswrong:
consider ops as classes. an incomplete abstract class is still a class

tqchen:
There are two semantics, here, both are valid, the difference is like follows

- Inheritate provides a ***pull semantics***, operators pull things from its parents,
    - There could be confusion about multiple inheritance in here
    - The recursive inheritance seems to be a overkill feature that is a bit hard to reason
    - It is not easy to tell from an abstract class from normal one, especially when multiple inheritance happen. 
- Match and set provides a ***push semantics***, which means find the ops that satisfies the condition and set the corresponding attribute.

Personally I prefer not introducing group as op, but simply as a returning result of some matching conditions on the operator set, which is aligned with push semantics. It also avoid introducing additional concept for specially handle groups.

tqchen:
@piiswrong changed API to include

src/core/op.cc
```diff
@@ -66,10 +72,42 @@ const any* Op::GetAttrMap(const std::string& key) {
...
+  auto& op_group = mgr->op_group;
+  for (const Op* op : dmlc::Registry<Op>::List()) {
+    if (op->index_ < op_group.size() &&
+        op_group[op->index_].count(group_name) != 0) {
+      trigger((Op*)op);  // NOLINT(*)
+    }
+  }
+}
+
+Op& Op::include(const std::string& group_name) {
```
piiswrong: document this?

tqchen: documents in the header

tqchen: https://github.com/dmlc/nnvm/pull/57/files#diff-c73a5e795727ea0120db27bc6d1b9bd1R234

tqchen: https://github.com/tqchen/nnvm/blob/master/include/nnvm/op.h#L234

PR #58, opened 2016-09-24
Add shape backward inference

tqchen:


tqchen:
@piiswrong

jermainewang:
Is this for inferencing the node entries in backward propagation? I actually have a change that output a mapping from GradientPass and use it in the InferPass. This solves the problem I mentioned in #43 . I don't know whether this is the same problem here. Shall I PR my change right now?

jermainewang:
The related codes are here:
GradientPass code change: https://github.com/jermainewang/nnvm/blob/partition/src/pass/gradient.cc#L152

InferPass code change:
https://github.com/jermainewang/nnvm/blob/partition/src/pass/infer_shape_type.cc#L114

tqchen:
We discussed this alternative with @piiswrong , because such mapping requires the gradient graph to remain unchanged (or change the mapping along during rewriting) we end up deciding to go with this alternative approach for now

tqchen:
This is an slightly different approach, that documents the backward operators's correspondence with forward operators, so shape inference will work when these information are registered. However, if the graph attribute transformation problem is solved, we would prefer to use the additional link attribute approach since that is simpler

jermainewang:
I actually have this TODO in the code. This is related to the liveness of graph attributes. Another thing is my PartitionPass actually requires this information. Can I keep the information in the GradientPass in the future while using your code in the InferPass?

tqchen:
I think so. We need to have more discussion on interface on how to make these information transformation easy. It is good to see these usecases surface up

tqchen:
@piiswrong If you can do a review tonight, let us merge this and make changes upstream in mxnet

piiswrong:
lgtm

PR #77, opened 2016-11-18
[RFC] PassManager

jermainewang:
PLEASE NOT MERGE. THIS IS ONLY FOR QUICK REVIEW.

@tqchen @piiswrong @ZihengJiang 

This is a rudimentary interface of Pass and PassManager. Also how Gradient and InferShapeType will change accordingly. Please have a look.

After we fix the interface, I will create the `pass-manager` branch in nnvm for further development.

### DIFF
--- a/include/nnvm/graph.h
+++ b/include/nnvm/graph.h
@@ -29,7 +29,9 @@ class Graph {
   /*! \brief outputs of the computation graph. */
   std::vector<NodeEntry> outputs;
   /*!
-   * \brief attributes of a graph
+   * \brief Attributes of a graph
+   *  [DEPRECATED!] Please use the set/get interface for different attribute types.
+   *
    *  Note that attribute is shared pointer and can be shared across graphs.
    *
    *  It is highly recommended to keep each attribute immutable.
@@ -40,13 +42,17 @@ class Graph {
    */
   std::unordered_map<std::string, std::shared_ptr<any> > attrs;
   /*!
-   * \brief Get the immutable attribute from attrs.
+   * \brief [DEPRECATED!] Get the immutable attribute from attrs.
    * \param attr_name the name of the attribute
    * \return the reference to corresponding attribute
    * \tparam T the type of the attribute.
    */
   template<typename T>
   inline const T& GetAttr(const std::string& attr_name) const;
+
+  template<typename T>
+  inline T& GetAttr(const std::string& attr_name);
+
   /*!
    * \brief Get a move copy of the attribute, implement copy on write semantics.
    *  The content is moved if the reference counter of shared_ptr is 1.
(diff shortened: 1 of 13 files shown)


jermainewang:
Changes highlights:

New Pass Interface:
--------------------------
```c++
class Pass {
  void Setup();
  PassResult RunOnGraph(Graph src, const PasssArgument& args);
  void Finalize();
};
```
`Setup` and `Finalize` will only be called once when the pass is created and destroyed, while `RunOnGraph` may be called multiple times.

New Attribute Interfaces
-----------------------------
The old adhoc attribute interface is still kept. The categorized attributes are simply implemented by prefixing to distinguish them.
* `Set/GetGraphAttr`: All graph attributes are destroyed if a graph is changed by a pass.
* `Set/GetNodeAttr`: All node attribute keys are prefixed by `__node_`. They will *only* be kept if PassResult specifies the `node_remapping` field by a pass.
* `Set/GetNodeEntryAttr`: All node entry attribute keys are prefixed by `__entry_`. They will *only* be kept if PassResult specifies the `entry_remapping` field by a pass.
* `Set/GetPassAttr`: All pass attribute keys are prefixed by `__pass_`. They will never be destroyed unless the corresponding pass is destroyed.

PassManager Inferface
---------------------------
* `Enable`: Enable a pass during running.
* `EnforceOrder`: Specify the order of two passes.
* `SetPassArguments`: Set the arguments of pass running.
* `Run`: Run all enabled passes in the order that we've discussed about.

Welcome any comments!

ZihengJiang:
should we consider that one pass may depend on multiple other passes?  if this situation is often, maybe it's good to provide an interface like `PassManager& EnforceOrder(const std::string& prev, const std::vector<std::string>& succ)`.

also, I'm curious about the schedule rule of the pass manager, if A ->(depend on) C, and B -> (depend on) C, will the sort should be `C=>A=>B`/ `C=>B=>A` or `C=>A=>C=>B`/`C=>B=>C=>A`?

jermainewang:
You mean both A and B will change the graph?

ZihengJiang:
right

jermainewang:
If C is an immutable pass, it should be C->A->C->B and Enforce(A, B) should be called otherwise error will be thrown. I think a more problematic one is if A and B are both mutable passes, A provides `a`, B provides `b` and another pass C depends on `a` and `b`, what is the order? It seems that we have cyclic problem here since schedule either A or B will erase the other's attributes.

ZihengJiang:
Maybe it's too rough to delete all previous results after every pass. We can consider a way to preserve the results of previous passes by their dependency.

piiswrong:
Any updates on this?
I'm worried some changes will break mxnet. Could you open a new branch for this?

jermainewang:
Sorry, haven't updated this for a long time. It is already on a new branch and I will merge the master myself. It is not meant to be merged in short time. Just ask for advices for the changes.

Issue #73, opened 2016-10-14
Shape Inference with Hints for Gradient Ops


tqchen:
Currently the shape/type inference for gradient op requires the registry of FBackwardInGradIndex and FBackwardOutToInIndex. Both are a bit cumbersome to do so.

Here is a second way to solve this problem
## In NNVM
- Add an operator in host language identity_attr_as_rhs(data, rhs), where the output is identity as data, but the type is constrained to be same as rhs
- In gradient pass, optionally enable pass a function AttrConstraintHint(input, rhs), which the source package will return identity_attr_as_rhs. This hint is inserted for every output of gradient ops, if FInferShape is not defined for that op.
  - This will not be inserted if FInfershape is already defined.
- Enable an attribute FIgnoreInputs  to the op do not have dataflow dependency on the data, but simply takes the information as hint  
  - Related ops include ones_like, and zeros_like
  -  Change memory allocation to take this into account and not retain memory for those index that are ignored.
## In Host Package(TinyFlow/ MXNet)
-  In host package, register identity_attr_as_rhs as identity compute, note that it will turn into a NOP if inplace happens.
- Update the https://github.com/dmlc/mxnet/blob/nnvm/src/executor/inplace_addto_detect_pass.cc to allow backprop replacement of addto to pass through identity maps

tqchen:
related to #70 https://github.com/dmlc/nnvm/issues/52

tqchen:
@piiswrong @jermainewang

jermainewang:
I think another way to do this is having a simple macro system. As we've discussed, the problem of maintaining attributes in graph is that the `Symbol` object does not contain attributes. So when user have following codes:
```python
z = x + y
dx = tf.grad(z, [x])   # Need to apply NNVM's Gradient pass here
ret = tf.exp(dx) * 2 - 3
```
, the gradient information cannot be carried on by the symbol composition. I think the main problem is that we are mixing the front-end construction and graph passes here. Remember, in a compiler stages, there is **no** looking back (e.g. once you finish syntax parsing, you go to semantic parsing and you could not go back). The problem here is that we want to have NNVM processing (similar to semantic parsing) in the middle of syntax parsing, which is object to the principal of compiler stages.

If the `tf.grad` is just a *macro* node that will be expanded in NNVM's passes, then all the information could be kept. So the above code will correspond to following graph:
```
[x, y] -> op(+) -> z
[x, z] -> macro_op(grad) -> dx
[...]

PR #79, opened 2016-11-21
Enable optional dependency memory and attr hint

tqchen:


tqchen:
per https://github.com/dmlc/nnvm/issues/73

PR #83, opened 2016-11-21
Fix json parsing behavior

piiswrong:
@tqchen


jermainewang:
BTW, why SaveJSON returns an empty graph rather than the original graph?

src/core/symbolic.cc
```diff
@@ -10,7 +10,7 @@
 namespace nnvm {
 
 namespace symbol_constants {
-const char *kNamespaceSeparator = "_";
+const char *kNamespaceSeparator = "$";
```
tqchen: will this make things not backward compatible?

piiswrong: saved json will be upgraded
the only thing that's broken is list_attr(recursive=True)
I raised a warning when it's called and replaced it with attr_dict

piiswrong: If we really want compatibility we can just replace $ with _ in frontend (the reverse is not possible)
but I think using _ as separation is really a bad idea and should be banned

tqchen: As long as the front-end behavior is fixed, I am fine with this change. Consider use "/" instead for separator, which is more typical

piiswrong: I wanted to use something no one would use so that its safe to split on it

piiswrong: also this is not exposed in frontend

PR #84, opened 2016-11-22
enable shape inference with hint func

tqchen:


tqchen:
https://github.com/dmlc/nnvm/issues/73

PR #85, opened 2016-11-25
Remove backward index, use gradient guessing instead

tqchen:
Solves #73


tqchen:
@piiswrong Tested on tinyflow and mxnet-nnvm

src/pass/infer_shape_type.cc
```diff
@@ -74,29 +75,44 @@ Graph InferAttr(Graph &&ret,
...
-        CHECK_LE(in_grad.size(), fnode.source->num_outputs());
-        for (size_t i = 0; i < in_grad.size(); ++i) {
-          uint32_t eid = idx.entry_id(inode.inputs[in_grad[i]]);
+      // input gradient list
+      auto igrad = fgrad[fwd_ptr->op()](fwd_ptr, ograd);
+      const Op* backward_op = inode.source->op();
+      const Node* igrad_node = nullptr;
+      // Input gradient assignement
+      for (size_t i = 0; i < igrad.size(); ++i) {
+        if (igrad[i].node->op() == backward_op) {
```
piiswrong: what if Fgradient you the same op twice?

piiswrong: Does the error point to FGradient?
Is it possible to add a hint after gradianpass calls FGradient

tqchen: no, the error occurs in the inference. However, in all our cases they do not occur

piiswrong:
How about check for name instead. FGradient has to give different names to
the two ops right?

tqchen:
discussed per offline, we concluded that current version is fine. since in all cases usually one backward op charges gradient for all(or there will be two kinds of backward ops). An error will occur to ensure things don't fall apart silently when violation happens

PR #92, opened 2017-01-03
add FSetInputVariableAttrs

piiswrong:
@tqchen


tqchen:
I do not quite understand what this attribute is intended for. Can you paste a example code on the comments?

piiswrong:
see the money pr
Tianqi Chen <notifications@github.com>于2017年1月3日 周二上午12:30写道：

piiswrong:
Mxnet pr
eric xie <eric.jy.xie@gmail.com>于2017年1月3日 周二上午12:32写道：

tqchen:
I have read the MXNet PR but still does not get the it clearly. Can you add a few comments and explanations directly?

piiswrong:
basically it gives operators the chance to put attributes into their input
variables. For example put initializer specifier string into it

tqchen:
This is OK to merge after the previous comment is fixed

piiswrong:
I think this affects all input variables. automatic creation is here https://github.com/piiswrong/nnvm/blob/8f11736006d957e51af2357de7c1ec6b4340104f/src/core/symbolic.cc#L304 . FSetAutoCreatedVariableAttrs will be called on all inputs

tqchen:
My main concern is that the naming of this function is confusing.n Just from the name, I would interpret it as a setter, instead of an automatic triggering function that mutates the state of its input.

tqchen:
How about FSetInputVarAttrOnCompose?

PR #94, opened 2017-01-07
enhance shape inference. allow in complete shape

piiswrong:
@tqchen


src/pass/infer_shape_type.cc
```diff
@@ -190,7 +192,7 @@ NNVM_REGISTER_PASS(InferShape)
         std::move(ret), TShape(),
         "FInferShape", "shape_inputs", "shape_attr_key",
         "shape", "shape_num_unknown_nodes",
-        [](const TShape& s) { return s.ndim() == 0; },
+        [](const TShape& s) { return s.ndim() == 0 || s.Size() == 0; },
```
tqchen: as an after thought, s.Size() check is not necessary, as most TShape use ndim to indicate None.

piiswrong:
I generalized it to allow incomplete shape like (0, 100, 100) to handle
cases like (0, 100) + (100, 0)

PR #97, opened 2017-01-16
change infer attr to match by name

piiswrong:
@tqchen 
Necessary for using multiple _zeros as fgradient


tqchen:
I do not like match by name, since this usually will create loopholes for future bugs. Let us think about alternatives

tqchen:
In theory, zeros's shape should not get inferred by itself, but instead by the op that takes zeros as input so we don't need backward inference logic to be able to handle zeros except for the top-most one when calculating the gradient. This could be handled with zeros_like. 

Unless there is a strong example on why we need this, let us close this

piiswrong:
example
```
struct ElemwiseGradZeros {
  std::vector<nnvm::NodeEntry> operator()(const nnvm::NodePtr& n,
                                          const std::vector<nnvm::NodeEntry>& ograds) {
    std::vector<nnvm::NodeEntry> ret;
    for (index_t i = 0; i < n->num_inputs(); ++i) {
      nnvm::NodePtr p = nnvm::Node::Create();
      p->attrs.op = nnvm::Op::Get("_zeros");
      p->attrs.name = n->attrs.name + "_backward_" + std::to_string(i);
      p->control_deps.emplace_back(n);
      ret.emplace_back(nnvm::NodeEntry{p, 0, 0});
    }
    return ret;
  }
};
```

tqchen:
The backward inference logic will only work if the Op itself is marked as a BackwardOp, say FullcBackward. In this case, ```_zeros``` is NOT a BackwardOp, so the backward shape inference won't be invoked.

I you could simply return zeros, as forward operator. And hopefully, the next operator that consumes them will be able to perform backward inference to set its shape correctly

PR #101, opened 2017-01-26
added ARM compiliation support

arank:
Made a simple conditional in makefile to disable msse compile flag to support arm devices.

Set export ARM=1 to enable ARM support


tqchen:
it is actually not needed, I have directly removed the option from make

vchuravy:
For future reference the is also the option `USE_SSE` that is used for MXNet proper to turn of SSE on power and ARM.

PR #102, opened 2017-01-31
Remove dmlc headers, add optional DMLC_CORE_PATH variable for build path

cjolivier01:


Makefile
```diff
@@ -12,6 +12,12 @@ include $(config)
 export LDFLAGS = -pthread -lm
 export CFLAGS = -std=c++11 -Wall -O2 -Iinclude -fPIC
 
+ifdef DMLC_CORE_PATH
```
tqchen: allow dmlc_core path to be NONE, to be consistent with rest of config, see ADD_CFLAGS

cjolivier01: When I do this, it is literally checking for DMLC_CORE_PATH being equal to the string "NONE".  Is this the intent?  For example, code below executes unless I have the line DMLC_CORE_PATH=NONE in config.mk.  Is this what is desired?

ifneq ($(DMLC_CORE_PATH), NONE)
  CFLAGS += -I$(DMLC_CORE_PATH)/include
endif

tqchen: yes, just to be consistent with the other part of config, like ADD_CFLAGS. I

cjolivier01: Maybe I am misunderstanding something:  in config.mk, it has the string ADD_CFLAGS= with no string after it.  It is harmless for a blank ADD_CFLAGS to be expanded in Makefile, but this isn't the case with DMLC_CORE_PATH, since the include selection follows from whether DMLC_CORE_PATH override is defined.  This would mean that config.mk would need to read:

# the additional link flags you want to add
ADD_LDFLAGS=

# the additional compile flags you want to add
ADD_CFLAGS=

# path to dmlc-core module 
DMLC_CORE_PATH=NONE


and not:

# the additional link flags you want to add
ADD_LDFLAGS=

# the additional compile flags you want to add
ADD_CFLAGS=

# path to dmlc-core module 
DMLC_CORE_PATH=

tqchen: hmm, after double checked the path configuration in mxnet. I think the ifdef way is actually better. So let us go with the current approach

tqchen:
good to merge once travis pass

Makefile
cjolivier01: ok

PR #105, opened 2017-02-24
LSTM Memory Allocator Fix #5035

eric-haibin-lin:
https://github.com/dmlc/mxnet/issues/5035


src/pass/plan_memory.cc
```diff
@@ -79,14 +87,16 @@ class GraphAllocator {
   }
 
   // constructor
-  explicit GraphAllocator(const IndexedGraph* idx) : idx_(idx) {
-    this->Init(dmlc::GetEnv("NNVM_EXEC_MATCH_RANGE", 16),
```
piiswrong: why remove NNVM_EXEC_MATCH_RANGE? what strategy are you using instead.

src/pass/plan_memory.cc
```diff
@@ -98,6 +108,21 @@ class GraphAllocator {
       num_match_color_ = pass::ColorNodeGroup(
           *idx_, importance, num_match_color_, &node_color_);
     }
+    // Init the free memory pool
+    if (shared_pool != nullptr) {
```
piiswrong: consider another strategy where you don't insert shared pool to free memory, instead you run planing separately, sort both shared_pool and new pool, then match from small to big.

tests/python/test_graph.py
```diff
@@ -130,6 +129,12 @@ def test_plan_memory():
     y = sym.add(y, y)
     g = graph.create(y)
     g._set_json_attr("shape_attr_key", "shape")
+    reused_bytes = 16
```
piiswrong: what are these?

src/pass/plan_memory.cc
eric-haibin-lin: I could try this

tests/python/test_graph.py
eric-haibin-lin: Maybe it's not clear. This is a test case where in the shared pool, there're 16 bytes on device 0, another 1 byte on device 0, and 32 bytes on device 9. Without the shared pool it takes 64 bytes. With this shared pool, total allocated mem should be 64 + 32 = 96 bytes for this graph. Should I test in finer granularity??

src/pass/plan_memory.cc
eric-haibin-lin: I'm trying to sweep between [1, NNVM_EXEC_MATCH_RANGE], and select the best value

tests/python/test_graph.py
piiswrong: No this is ok

src/pass/plan_memory.cc
```diff
@@ -197,28 +166,38 @@ Graph PlanMemory(Graph ret) {
...
             ref_count[eid_out] != 0 &&
-            storage[eid_out] == GraphAllocator::kBadStorageID &&
-            storage[eid_in] != GraphAllocator::kBadStorageID &&
+            (*storage)[eid_out] == GraphAllocator::kBadStorageID &&
+            (*storage)[eid_in] != GraphAllocator::kBadStorageID &&
             shape_vec[eid_out].Size() == shape_vec[eid_in].Size() &&
             dtype_vec[eid_out] == dtype_vec[eid_in]) {
           // inplace optimization
-          storage[eid_out] = storage[eid_in];
+          (*storage)[eid_out] = (*storage)[eid_in];
```
tqchen: consider add a reference in to *storage in beginning?

src/pass/plan_memory.cc
```diff
@@ -197,28 +166,38 @@ Graph PlanMemory(Graph ret) {
...
-        storage[eid] = allocator.Request(dev_id, dtype_vec[eid], shape_vec[eid], nid);
+      if ((*storage)[eid] == GraphAllocator::kBadStorageID) {
+        auto &eshape = shape_vec[eid];
+        size_t esize = 0;
+        if (eshape.ndim() != 0) esize = eshape.Size();
+        eids.insert(std::make_pair(esize, eid));
       }
     }
+    std::multimap<size_t, uint32_t>::reverse_iterator rit;
+    for (rit = eids.rbegin(); rit != eids.rend(); ++rit) {
```
tqchen: auto rit = eids.rbegin()

src/pass/plan_memory.cc
eric-haibin-lin: I actually started with reference. But python lint gives a complaint that i either pass reference variable which i don't modify, or pass a pointer to a variable which mutate. I don't remember if it's enforced in dmlc/mxnet or dmlc/nnvm.

src/pass/plan_memory.cc
eric-haibin-lin: will do

src/pass/plan_memory.cc
eric-haibin-lin: `src/pass/plan_memory.cc:144:  Is this a non-const reference? If so, make const or use a pointer: StorageVector& storage  [runtime/references] [2]`
This is the error message when I pass as reference

tqchen: You can pass by pointer but make reference inside the function.

PR #107, opened 2017-03-09
Enable shape hints during infer_shape pass

ZihengJiang:
This is a temporary solution to solve the dependency between backward_op and forward_op in infer_shape pass, which will enable users to give shape hints from outside. A refactor of this part is expected in the near future.  @piiswrong


include/nnvm/node.h
```diff
@@ -41,6 +41,30 @@ struct NodeEntry {
...
+struct NodeEntryEqual {
+  size_t operator()(const NodeEntry& a, const NodeEntry& b) const {
+    return (a.node.get() == b.node.get()) &&
+           (a.index == b.index) &&
+           (a.version == b.version);
+  }
+};
+
+/*! use a NodeEntry as a key in unordered_map */
+template<typename Foo>
```
jermainewang: What is `Foo`?

include/nnvm/node.h
```diff
@@ -41,6 +41,30 @@ struct NodeEntry {
   uint32_t version;
 };
 
+/** This lets you use a NodeEntry as a key in a unordered_map of the form
+ * unordered_map<NodeEntry, Foo, NodeEntryHash, NOdeEntryEqual> */
```
jermainewang: Comment format.

pluskid:
lol, I like `typename Foo`.

PR #108, opened 2017-03-25
Add codegen field in graph node

jermainewang:
Motivation Issue
------------------
During autograd, we first generate forward path by concrete execution. We then call `GradientPass` to generate backward graph. We then **prune out** the forward graph since it has already been executed and no resources should be allocated as well. However, this approach fails in https://github.com/dmlc/mxnet/blob/master/src/executor/attach_op_execs_pass.cc#L202 . This is because converting BackwardOpExecutor relies on forward node, which has already been deleted.

There are two ways to view this problem:
* We should not remove the forward nodes, since some of their states should be preserved (i.e, it's the pass developer's fault).
* Generating executors (codes) for one node should not rely on other nodes, since they may be deleted (i.e, it's the mxnet backend developer's fault).

I think both views are somewhat correct, and the problem is there is **no clear semantics** on what is a valid IR (dataflow graph) that can be accepted. If take LLVM as analogy, LLVM pass does not allow to generate IR that is invalid (based on its definition). Here in NNVM, this part is missing.

Defining Semantics of NNVM IR
------------------------------------
The reason is because the node is basically a black box. However, we can still define some high-level semantics of the node. Therefore, I suggest to add a `CodeGen` field in the `Node` structure meaning that **the node is able to convert to some codes for the backend**.

A dataflow graph (NNVM IR) is valid *if and only if* all the nodes in the graph has defined code generator. The code generator has following interface:
```c++
class CodeGen {
 public:
  virtual void FeedInputShapes(const std::vector<TShape>& shapes) = 0;
  virtual void FeedOutputShapes(const std::vector<TShape>& shapes) = 0;
  virtual void FeedInputTypes(const std::vector<int>& types) = 0;
  virtual void Generate(const Node* node) = 0;
};
```
Here, it relies *only* on the current node to generate code (or OpExecutor in MXNet's backend). Each pass must generate valid IR. Otherwise, error is raised.

How this decouples forward & backward operators
--------------------------------------------------
Based on the above definition, now it is the `GradientPass` 's responsibility to generate valid IR so that each node has valid CodeGen. The backward node's CodeGen is created in this pass by storing the `NodePtr` directly to the CodeGen object. This allows succeeding passes to remove any nodes since the required states have been stored. Two op attributes are added to support this:
```c++
using FCodeGenCreator = std::function<std::shared_ptr<CodeGen>()>;

using FBackwardCodeGenCreator = std::function<std::shared_ptr<CodeGen>(
    const NodePtr& fwd_node)>;
```
I have implemented the mechanism in my repo [here](https://github.com/jermainewang/mxnet/blob/master/src/nnvm/legacy_op_util.cc#L524). This also greatly simplified the `attach_op_exec_pass` (i.e, the code gen part in MXNet backend). The demo code is [here](https://github.com/jermainewang/mxnet/blob/master/src/executor/attach_op_execs_pass.cc#L63). You can see creating `OpExecutor` for backward node no longer depends on forward node.

Q: *We already have `Op` pointer. Why another `CodeGen` field?*
A: The `Op*` field is a static attribute of the node. For example, two conv nodes will share the `Op*` while they may have different code gen.

If you guys think this is a necessary change. I will keep working on this including:
* Replace the `NDArrayFunction` with `CodeGen`.
* Replace the `FCompute` with `CodeGen`.
So everything will become `CodeGen`.

@tqchen @piiswrong @ZihengJiang

### DIFF
--- a/include/nnvm/codegen.h
+++ b/include/nnvm/codegen.h
@@ -1,0 +1,21 @@
+#ifndef NNVM_CODEGEN_H_
+#define NNVM_CODEGEN_H_
+
+#include <memory>
+#include <vector>
+#include "./tuple.h"
+#include "./node.h"
+
+namespace nnvm {
+
+class CodeGen {
+ public:
+  virtual void FeedInputShapes(const std::vector<TShape>& shapes) = 0;
+  virtual void FeedOutputShapes(const std::vector<TShape>& shapes) = 0;
+  virtual void FeedInputTypes(const std::vector<int>& types) = 0;
+  virtual void Generate(const Node* node) = 0;
+};
+
+}  // namespace nnvm
+
+#endif  // NNVM_CODEGEN_H_
--- a/include/nnvm/node.h
+++ b/include/nnvm/node.h
@@ -17,6 +17,7 @@ namespace nnvm {
 
 // Forward declare node.
 class Node;
+class CodeGen;
 
 /*!
  * \brief we always used NodePtr for a reference pointer
@@ -91,6 +92,7 @@ struct NodeAttrs {
    * The object can be used to quickly access attributes.
    */
   any parsed;
+
 };
 
 /*!
@@ -107,6 +109,11 @@ class Node {
    *  Gives operation must be performed before this operation.
    */
   std::vector<NodePtr> control_deps;
+
+  // Can only be null if its op* is also null, which means it is a
+  // variable node.
+  std::shared_ptr<CodeGen> codegen{nullptr};
+
   /*! \brief destructor of node */
   ~Node();
   /*! \return operator in this node */
--- a/include/nnvm/op_attr_types.h
+++ b/include/nnvm/op_attr_types.h
@@ -12,6 +12,7 @@
 #include <functional>
 #include "./base.h"
 #include "./node.h"
+#include "./codegen.h"
 #include "./tuple.h"
 
 namespace nnvm {
@@ -153,6 +154,11 @@ using FSetInputVarAttrOnCompose = std::function<void(
     const NodeAttrs& attrs,
     NodePtr var,
     const int index)>;
+
+using FCodeGenCreator = std::function<std::shared_ptr<CodeGen>()>;
+
+using FBackwardCodeGenCreator = std::function<std::shared_ptr<CodeGen>(
+    const NodePtr& fwd_node)>;
 
 }  // namespace nnvm
 
--- a/src/core/symbolic.cc
+++ b/src/core/symbolic.cc
@@ -23,6 +23,7 @@ NodePtr CreateVariableNode(const std::string& name) {
   n->attrs.op = nullptr;
   n->attrs.name = name;
   n->attrs.parsed = VariableParam();
+  n->codegen = nullptr;
   return n;
 }
 
@@ -509,6 +510,7 @@ std::vector<std::tuple<std::string, std::string, std::string> >
 Symbol Symbol::CreateFunctor(const Op* op,
                              std::unordered_map<std::string, std::string> attrs) {
   static auto& fnum_vis_output = Op::GetAttr<FNumVisibleOutputs>("FNumVisibleOutputs");
+  static auto& fcodegen = Op::GetAttr<FCodeGenCreator>("FCodeGenCreator");
   Symbol s;
   NodePtr n = Node::Create();
   n->attrs.op = op;
@@ -516,6 +518,10 @@ Symbol Symbol::CreateFunctor(const Op* op,
   if (n->op()->attr_parser != nullptr) {
     n->op()->attr_parser(&(n->attrs));
   }
+  CHECK(fcodegen.count(op))
+    << "No code generator is registered for operator \""
+    << op->name << "\"";
+  n->codegen = fcodegen[op]();
 
   uint32_t nout = n->num_outputs();
   if (fnum_vis_output.count(n->op())) {


jermainewang:
Also mentioned that, @ZihengJiang and I have discussion on another solution: partial execution to solve the forward/backward operator issues. The idea is to keep the whole graph but only execute part of the nodes. However, this seems to require more changes since we need to basically change all passes to avoid allocating resources to those nodes that are not executed.

piiswrong:
The forward operator shouldn't have been deleted. If it's deleted there is
no way to recover

Minjie Wang <notifications@github.com>于2017年3月24日 周五下午8:10写道：

jermainewang:
We think the executor should have no semantics of a neural network (thus the so-called forward/backward concepts). By storing the `NodePtr` in the `CodeGen` object, deleting forward node is safe while not violating the IR.

piiswrong:
Operator objects can have states. If you delete it  the states will be
permanently lost. There is no way to safely delete it.

Minjie Wang <notifications@github.com>于2017年3月24日 周五下午8:14写道：

jermainewang:
You can read my implementation. The forward and backward shared `CodeGen` operator so even if the forward node is deleted, the `OpExecutor` can still be created.

jermainewang:
The `Operator` are created from `NodeAttrs`. If `NodeAttrs` are kept, they everything can be recovered.

piiswrong:
The operator object can have states produced by forward and used by
backward. It can not be recovered from nodeattr

Minjie Wang <notifications@github.com>于2017年3月24日 周五下午8:21写道：

jermainewang:
In my view, this should not be allowed. All members in the operator should be constants after `Init`. Developers should pass these intermediate data through the output and feed it to backward through inputs.

piiswrong:
It's has always been supported. We shouldn't break backward compatibility.


Minjie Wang <notifications@github.com>于2017年3月24日 周五下午8:28写道：

jermainewang:
Do you have any examples?

piiswrong:
Cudnn ops mkl ops custom op pretty much every non trivial op

Minjie Wang <notifications@github.com>于2017年3月24日 周五下午8:30写道：

jermainewang:
We've checked the cudnn operators. They are fine since you can still create the operator and only call backward. The only problem it wastes some storage for the forward handle, but it should be fine.

piiswrong:
A lot more ops keep private states. Custom op  warpctc torchop caffeop etc

Minjie Wang <notifications@github.com>于2017年3月24日 周五下午8:33写道：

jermainewang:
Actually, if this is really a crucial use case. They should be represented as stateful nodes in the dataflow graph that can have multiple calls and are serialized. Basically, we should have a systematic way to tell the pass developer that "ok, here are some nodes are very dangerous and please do not break their relations". But simply mark them as forward/backward is not a good idea i think.

jermainewang:
Also, I think those use cases can still be solved under the above change. When you creating the backward nodes, directly save the forward operator (that has already been executed) in the `CodeGen` object, so it can be reused later.

jermainewang:
It is similar to you want to "generate backward code that based on the forward running states". This concept should be fine.

piiswrong:
There is no need for FeedInputShapes etc since they are not sufficient for recovering forward state anyway.

CreateLayer should directly save created operator in NodeAttrs. Maybe a new dmlc::any field or reuse NodeAttrs::parsed.

jermainewang:
This is again not clear. I think `NodeAttrs` are static attributes before compilation. `Operator` is runtime object after running. They should not be put together.

piiswrong:
node is in theory static too. Only graph attributes are dynamic

tqchen:
There are a few problems with the codegen interface:
- It is restricted to shape, and type, and may not be future proof on other possible context that you might want to depend on(e,g. layout)
- It seems to be coupled with the execution, while the codegen itself is a blackbox.
- The stateful interface means it is not thread-safe.

I would try not to introduce the concept to core graph structure, but maybe start by thinking about how states could have been stored without directly think about execution. an anyfield might be useful for such case.

tqchen:
Maybe a clear state sharing mechanism is preferred than the implicit graph deps.  The key issue is that the layer operator with shared state is never able to do autograd(if the state is not recovered).

So I think we should simply mark these stateful operator as they are, and gradually move most operators to stateless(for example, cudnn conv could be stateless)  and only do autograd on stateless ops.

jermainewang:
The interface actually does not restrict the information required. Actually, in my implementation, I have a `MXNetCodeGen` that inhertits from this interface and add more interface: https://github.com/jermainewang/mxnet/blob/master/src/executor/exec_pass.h#L56 . The interface means it is totally fine these `FeedXXX` is not called. In fact, following interface is what I wrote initially:
```c++
class CodeGen {
 public:
  virtual void Generate(const Node* node) = 0;
};
```
I guess this is the "black-box" you want?

>The stateful interface means it is not thread-safe.
What do you mean by thread-safe? During execution or during compilation?

Just chatted with eric, we agree the state sharing interface is required.

tqchen:
The part that confuses me is that CodeGen seems to be a mutable object, in a graph with immutable attributes.

tqchen:
Capturing forward nodeptr in the codegen or some object seems did not solve the problem of the state sharing, unless forward computation also uses a similar mechanism. I think it might be helpful to have a clear specification of shared states.

On the other hand, autograd can already cover most of the ops given that they do not share any states, which should be the case actually for almost all of them. For temporal objects such as resource handle, a thread local workspace or an explicit resource should be likely solve the problem

jermainewang:
I see what you mean. Autograd allows compute gradients of symbols together with ndarrays. Unfortunately, symbols usually represent stateful ops (e.g. conv). Another thing is, the `Operator` interface right now does not allow creating the `OpExecutor` with only backward node. Therefore, even if they are stateless, the backward `OpExecutor` cannot be even created (or you still view them as stateful since their  code generation is also shared?).

tqchen:
If operator is stateless, we can simply create a OpExecutor(which is forward) that does the backward computation

jermainewang:
@tqchen  What did you mean by " For temporal objects such as resource handle, a thread local workspace or an explicit resource should be likely solve the problem" ?

jermainewang:
Just discussed with tianqi offline. Conclusion: we will walkaround this by modifying those legacy operator interface. We need to expose two more interfaces on the legacy `Operator`:
* Whether they are stateless or not?
* For stateless operator, how to create the backward OpExecutor given its input and output shape/type (no need for forward node's shape/type)?

For stateful operator, we just ignore them in autograd right now.
@ZihengJiang @piiswrong 

If guys think this is a good walkaround right now. I will close this PR and raise another PR to MXNet repo about the change.

piiswrong:
operators should be stateful by default. Because users self defined
operators don't have this interface defined

Minjie Wang <notifications@github.com>于2017年3月25日 周六上午10:49写道：

piiswrong:
Also this is cluttering the interface. Stateless ops should be implemented
with fcompute there is no need to label legacy operators stateless.

eric xie <eric.jy.xie@gmail.com>于2017年3月25日 周六下午10:15写道：

jermainewang:
Reopen for discussion. The thing is most current money op are in legacy operator interface. Yes, the interface allows states but in fact most of them are stateless. Currently, we don't have a clean way to specify state sharing. So if we treat them all stateful, autograft will just don't allow them and all the symbol part can not be supported. That's why we need to add two simple interfaces, which by default return stateful so nothing will change. And gradually mark those stateless operators. It will not influence any of the current mechansim and is backward compatible. Is this OK for you?

jermainewang:
If operator is stateless, then we can either convert it to fcompute registration or create a new operator for backward so it does not need to refer to forward node in attach op exec pass.

PR #109, opened 2017-03-30
change to match by both op and name

piiswrong:


tqchen:
Why do we need match by name?

piiswrong:
it's safer. Some times you can use the same op for different ingrad

tqchen:
I assume it is not safer, since there is no requirements in the naming spec. Supporting this will enable ops that relies on this feature(which is not very reliable) we should not do this unless there is a concrete op that does so

piiswrong:
reverted.
please have a look at the rest

PR #113, opened 2017-05-03
force infer storage on backward pass (#1)

eric-haibin-lin:
This PR allows inferstoragetype in backward pass. For type and shape, we can skip backward inter pass because it's always identical, but it's not the case for storage type (e.g. For sparse embedding operator, we'll output RowSparse weight gradients).

### DIFF
--- a/src/core/symbolic.cc
+++ b/src/core/symbolic.cc
@@ -436,7 +436,6 @@ Symbol Symbol::GetInternals() const {
 }
 
 Symbol Symbol::GetChildren() const {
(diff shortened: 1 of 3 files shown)


eric-haibin-lin:
@piiswrong merging this helps me enable some CI tests on PR https://github.com/dmlc/mxnet/pull/5800

piiswrong:
test failed

eric-haibin-lin:
@piiswrong updated unit test

PR #114, opened 2017-05-16
Rename TShape::index_t to dim_t and change to int64_t

ZihengJiang:


include/nnvm/tuple.h
```diff
@@ -535,6 +528,55 @@ class TShape : public Tuple<index_t> {
...
+/*! \brief helper function to transform a container to TShape with type cast */
+template<typename SrcIter>
+inline TShape ShapeTypeCast(const SrcIter begin, const SrcIter end) {
+  size_t ndim = std::distance(begin, end);
+  TShape res(ndim);
+  ShapeTypeCast(begin, end, res.begin());
+  return res;
+}
+
+template<typename ValueType>
```
tqchen: add comment about what is value type

include/nnvm/tuple.h
```diff
@@ -381,15 +374,15 @@ class TShape : public Tuple<index_t> {
    * \param src source shape.
    * \return self.
    */
-  inline TShape& operator=(Tuple<index_t>&& src) {  // NOLINT(*)
+  inline TShape& operator=(Tuple<dim_t>&& src) {  // NOLINT(*)
     TShape(std::move(src)).swap(*this);  // NOLINT(*)
     return *this;
   }
   /*! \return total number of elements in the shape */
   inline size_t Size() const {
```
eric-haibin-lin: Is size_t sufficient to hold the result here?

ZihengJiang: It should be enough in normal situation

PR #121, opened 2017-06-22
delete dependency node

formath:
control node should be deleted explicitly like that of input node? @piiswrong


piiswrong:
@tqchen

PR #122, opened 2017-06-23
Handling duplicate NodeEntries on the edge of the gradient graph

ptrendx:
In  this PR I check for duplicates in Gradient graph outputs and (if the copy operation is provided) make new NodeEntries that are copies of the duplicates (effectively ensuring that all outputs from gradient pass are unique).

### DIFF
--- a/src/pass/gradient.cc
+++ b/src/pass/gradient.cc
@@ -220,7 +220,7 @@ Graph Gradient(Graph src) {
         unique_grads.emplace(NodeEntry{std::move(copy_node), 0, 0}, std::make_pair(1, counter));
       }
     } else {
-        ret.outputs.emplace_back(std::move(entry.sum));
+        ret.outputs[counter] = entry.sum;
     }
     ++counter;
   }
--- a/src/pass/gradient.cc
+++ b/src/pass/gradient.cc
@@ -91,6 +91,9 @@ Graph Gradient(Graph src) {
   if (src.attrs.count("zero_ops") != 0) {
     zero_ops = src.GetAttr<std::vector<const Op*> >("zero_ops");
   }
+  const Op* copy_op = (src.attrs.count("copy_op") != 0) ?
+      Op::Get(src.GetAttr<std::string>("copy_op")) :
+      nullptr;
 
   // topo sort
   std::vector<NodePtr> topo_order;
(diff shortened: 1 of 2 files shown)


piiswrong:
@tqchen

src/pass/gradient.cc
```diff
@@ -200,7 +205,29 @@ Graph Gradient(Graph src) {
...
+        copy_node->inputs.emplace_back(entry.sum);
+        unique_grads.emplace(NodeEntry{std::move(copy_node), 0, 0}, std::make_pair(1, counter));
+      }
+    } else {
+        ret.outputs[counter] = entry.sum;
+    }
+    ++counter;
+  }
+  if (copy_op != nullptr) {
+    for (auto kv : unique_grads) {
```
tqchen: const auto&

tqchen:
besides my comment, other part lgtm

src/pass/gradient.cc
```diff
@@ -200,7 +205,29 @@ Graph Gradient(Graph src) {
...
+        unique_grads.emplace(NodeEntry{std::move(copy_node), 0, 0}, std::make_pair(1, counter));
+      }
+    } else {
+        ret.outputs[counter] = entry.sum;
+    }
+    ++counter;
+  }
+  if (copy_op != nullptr) {
+    for (const auto& kv : unique_grads) {
+      ret.outputs[kv.second.second] = kv.first;
```
piiswrong: Would this permute the order of returned gradients? Is this a problem? @tqchen

piiswrong: never mind.

src/pass/gradient.cc
```diff
@@ -200,7 +205,29 @@ Graph Gradient(Graph src) {
...
+    if (copy_op != nullptr) {
+      auto kv = unique_grads.find(entry.sum);
+      if (kv == unique_grads.end()) {
+        unique_grads.emplace(std::move(entry.sum), std::make_pair(1, counter));
+      } else {
+        NodePtr copy_node = Node::Create();
+        std::ostringstream os;
+        os << entry.sum.node->attrs.name << "_" << kv->second.first << "_copy";
+        kv->second.first++;
+        copy_node->attrs.op = copy_op;
```
piiswrong: if attr_parse != nult attr_parser()

src/pass/gradient.cc
ptrendx: No - kv.second.second stores the original ordering - it would be permuted otherwise (and would not work because of that).

src/pass/gradient.cc
ptrendx: Something like this (from place_device pass)?
```
            if (copy_node->attrs.op->attr_parser != nullptr) {
              copy_node->attrs.op->attr_parser(&(copy_node->attrs));
            }
```

piiswrong: yes

PR #124, opened 2017-06-29
fix plan memory add inplace_identity option

piiswrong:


include/nnvm/op_attr_types.h
```diff
@@ -118,6 +116,18 @@ using FInplaceOption = std::function<
...
+ * \brief Get if the inplace option is an identity
+ *  This function enables inplace optimization even when input reference count
+ *  is greater than one.
+ * \param attrs The attributes of the node
+ * \return list of bool indicating whether corresponding pair from FInplaceOption
+ *         is an identity
+ *
+ * \note Register under "FInplaceIdentity", by default no identities.
+ */
+using FInplaceIdentity = std::function<std::vector<bool> (const NodeAttrs& attrs)>;
```
tqchen: maybe only specialize to single in-out case?

src/pass/plan_memory.cc
```diff
@@ -152,31 +156,46 @@ size_t AllocMemory(const Graph& ret, const IndexedGraph& idx, StorageVector* sto
...
+            (storage_ref_count[sid_in] == 1 || identity[ipair]) &&
+            entry_ref_count[eid_out] > 0 &&
             shape_vec[eid_out].Size() == shape_vec[eid_in].Size() &&
             dtype_vec[eid_out] == dtype_vec[eid_in]) {
           // inplace optimization
-          storage[eid_out] = storage[eid_in];
-          ref_count[eid_in] = 0;
+          taken[kv.first] = true;
+          storage[eid_out] = sid_in;
+          storage_ref_count[sid_in] += entry_ref_count[eid_out];
```
tqchen: It would be very helpful to mark where storage_ref_count substraction happens to substract this to 0, especially when there is an identity pair. Would be nice to write some comment on how the algorithm works

include/nnvm/op_attr_types.h
tqchen: is it sufficient to consider single in/out case?

PR #126, opened 2017-07-12
Fixed make no_optimization runs indefinitely when run the memcost example provided 

zhuhyc:
Fixed make no_optimization runs indefinitely when run the memcost example provided

Run the memcost example provided [here](https://github.com/dmlc/mxnet/tree/master/example/memcost).

    make with_inplace, make with_sharing and make with_both all three use the same amount of Memory Total 1340 MB allocated.

    make no_optimization runs indefinitely
Environment info

MXNet versions tested:

    0.10.1
    build from source / master

Operating System: Linux


zhuhyc:
@tqchen ping

PR #132, opened 2017-07-28
[ATTR/SYMBOL] Expose op_name attr to python

tqchen:


tqchen:
It is useful to do reflection and graph transformation in the python side

tqchen:
@piiswrong

PR #140, opened 2017-08-19
[WIP] Build nnvm as a static lib instead of a shared lib.

weijianwen:
This commit enable cmake to build and install NNVM static lib and header files, which can be reused by other libraries via `include` and `lib`. Compared to a shared one, a static lib if preferred when bulidng MXNet. See `$(NNVM_PATH)/lib/libnnvm.a` in https://github.com/apache/incubator-mxnet/blob/master/Makefile .


piiswrong:
@cjolivier01

cjolivier01:
Would it be possible to please make this an option (ie USE_STATIC_NNVM or something like that?)?  This makes pip distribution a bit more messy.  Also, can you please check that mxnet still builds everything, including im2rec, mxnet_unit_tests and image-classification-predict?  These tend to be sensitive to whether dependent libraries are static or dynamic (the wrong combination can cause some portions of the static init to run twice).  Note that currently in mxnet, I believe we are using -Wl,--whole-archive for nnvm static in order to force inclusion of all static initialization.

cjolivier01:
Note in mxnet CMakeLists.txt, I have a section that creates bith a static and dynamic library
There's a better way for doing this with cmake (something like TARGET:objs), but that doesn't work with the cuda objects.

  if(UNIX)
    add_library(mxnet_static STATIC ${SOURCE})
    # Need an arbitrary source file to trigger CMake to build the library
    add_library(mxnet SHARED)
    # This has prolems, as it adds libmxnet_static to INTERFACE_LINK_LIBRARIES
    target_link_libraries(mxnet "-Wl,--whole-archive $<TARGET_FILE:mxnet_static> -Wl,--no-whole-archive")
    target_link_libraries(mxnet mxnet_static) # Let cmake understand the dependency
    add_custom_target(
        StaticallyLinkStaticMXNetLibrary ALL
        BYPRODUCTS ${CMAKE_ARCHIVE_OUTPUT_DIRECTORY}/libmxnet.a
        WORKING_DIRECTORY ${CMAKE_ARCHIVE_OUTPUT_DIRECTORY}
        COMMAND ln -sf libmxnet_static.a libmxnet.a
        DEPENDS mxnet_static
    )
  else()
    add_library(mxnet SHARED ${SOURCE})
  endif()

weijianwen:
I followed MXNet's official installation guide in which `Makefile` is used rather than `CMakeLists.txt`. After switching nnvm to a static build, everythig in MXNet has been built successfully. I didn't try building MXNet itself with CMake, but I built all its dependency packages (ps-lite, nnvm, mshadow...) with CMake. Just hope that one day we can eliminate `Makefile`.

@cjolivier01  Do you suggest adding a flag into nnvm's CMake file to control whether or not building a static library? From your experiences, do you have any concern on CMake's machenism to generate a static library and how I can avoid it. Thank you.

cjolivier01:
does mxnet_unit_tests also run and pass everything?
whatever output created by Makefile should also be created by cmake.
If all that is the case, then I am ok.

weijianwen:
Hmm. I run the MNIST tests but not MXNet's unit tests. I marked this commit as work in progessn now. I plan to:

1. Update to add `BUILD_STATIC` (`NO` by default) and `BUILD_SHARED` (`YES` by default).
2. Suggested by @cjolivier01 , running unit tests in MXNet against the static `libnnvm` library .

The default behavior of CMake in nnvm will NOT change, just building shared library as before.

cjolivier01:
The reason I ask about the mxnet_unit_tests is just because of the double-init problem that I have seen. If it doesn't blow up immediately, then it's ok.

weijianwen:
@cjolivier01 could you check #146 at your earliest convenience? This PR is closed by accident and github refuses to reopen it.

cjolivier01:
It seems you can't reopen it because the branch is deleted

PR #145, opened 2017-08-27
add additional field into node for autograd

piiswrong:


jermainewang:
I feel like there is a growing requirement for row-stored attribute in graph (just like your proposal). Could we discuss more on this? For example, will `map<string, shared_ptr<any>>` be a better abstraction?

piiswrong:
You can put map<string, shared_ptr<any>> into any

PR #151, opened 2017-09-22
declare a type name for each tuple type

pierric:
a generic way to declare a type-name for Tuple


pierric:
travis fails due to a reference a latest commit to dmlc_core, which itself is in a pull request. 
I have add in a type_name_helper class to replace the type_name(), because as a function type_name() cannot be partially specialized for Tuple<T>. I can resolve the error once after the pull request in dmlc_core is merged.

PR #152, opened 2017-09-25
[JSON] attr->attrs in node attribute, keep backward read compatible

tqchen:


tqchen:
@piiswrong

piiswrong:
why?

tqchen:
To be consistent with the graph attributes(which is named as attrs) and other naming(e.g. inputs instead of input)

tqchen:
@piiswrong followup on this?

piiswrong:
what graph attributes?

tqchen:
When we save graph, the graph attributes(shape type) are stored as attrs instead of attr

PR #153, opened 2017-09-27
Address execution hot spots

cjolivier01:
Copies while resizing on push_back() is degrading performance. Calling reserve() allows enough space in the array for the maximum number of items which may be inserted.


cjolivier01:
Ok, ignore this PR, got some noise in here somehow

PR #154, opened 2017-09-27
Address execution hot spots

cjolivier01:
resize+copy when adding items to these vectors are witnessed performance problems.  reserve() pre-allocates memory for the results.


tqchen:
I think these are good for temporary vectors. For the thread local vector, however, the impact is not as large. Because the vector's internal space is reserved across multiple calls

cjolivier01:
Which item are you referring to? The ones which used to call clear()? Depending upon the implementation, clear() may free the memory. resize(0) is guaranteed to retain the memory. Both of these types of called showed up as hot spots in a vtune analysis.

tqchen:
I see, I thought clear won't free up memory

cjolivier01:
Depends on the implementation. I can use clear() and reserve instead of reaize(0) if you like.

tqchen:
I think resize(0) is fine, since the indention is to not free memory anyway

PR #173, opened 2017-10-15
[TVM] upgrade to generic schedule

tqchen:


tqchen:
@ZihengJiang please verify that rasp schedule still works correctly

Issue #180, opened 2017-10-18
problem with coreml model on nnvm


gengtsh:
when i run my coreml on nnvm, i encountered one issue:
AttributeError: 'module' object has no attribute 'concat'
then i check symbol.py in nnvm/python/nnvm folder, there is no concat definition in Symbol class.
i notice concat is defined in onnx.py, so, do i need to import onnx by myself? or what can i do it for this issue? thanks a lot.

PENGUINLIONG:
Both `Concat` and `concat` can be used to make concatenation layers in mxnet but for coreml it's not the case. There is [`ConcatLayerParams` for concatenation](https://github.com/dmlc/nnvm/blob/74549d68593115cf59492d0d7852f06223d697e6/python/nnvm/frontend/coreml.py#L286) in `coreml.py` yet it has another name. What defined in `onnx.py` is for onnx, they are not the same thing. Both `Concat` and `concat` can be used to make concatenation layers in mxnet but for coreml it's not the case.

gengtsh:
could you please explain a little about your first sentence "...for coreml it's not the case".  does it mean that coreml didn't support concat ? 

my problem is:
my original model is a caffe model. I  use coremltools, convert caffe model to coreml model, then when i run it on the nnvm, it said that" AttributeError: 'module' object has no attribute 'concat'" . 
The problem is in line 251 (ret = _sym.concat(ret, insyms[i], dim=1) ) in coreml.py. it is in def ConcatLayerParams. based on my understanding, _sym.concat is not implemented in symbol.py. that's the reason why i get this issues.  if i'm wrong, please correct me. 

could you please tell me how to handle this issue? thanks in advance.

PENGUINLIONG:
I see the problem. The concatenation op is registered as [`concatenate`](https://github.com/dmlc/nnvm/blob/a96133e4074e61054d106efab2ca4ebebbde7454/python/nnvm/top/transform.py#L49-L57) but is not used as-is. It's called `concat` in [`coreml.py`](https://github.com/dmlc/nnvm/blob/74549d68593115cf59492d0d7852f06223d697e6/python/nnvm/frontend/coreml.py#L251) instead of the correct one, `concatenate`. Rename it and it should work fine. I will PR that.

tqchen:
thanks @PENGUINLIONG for fixing this issue

PR #181, opened 2017-10-18
[FRONTEND] Correct the use of `concatenate` operator

PENGUINLIONG:
See issue #180 .


python/nnvm/frontend/coreml.py
```diff
@@ -248,7 +248,7 @@ def ConcatLayerParams(op, insyms, symtab):
         raise NotImplementedError("Sequence Concat not supported")
     ret = insyms[0]
     for i in range(1, len(insyms)):
-        ret = _sym.concat(ret, insyms[i], dim=1)
+        ret = _sym.concatenate(ret, insyms[i], dim=1)
```
tqchen: we can do ret = _sym.concatenate(*insyms, dim=1)

PENGUINLIONG: There are also some `ret`s used as parameters nearby (like [this one](https://github.com/dmlc/nnvm/blob/a96133e4074e61054d106efab2ca4ebebbde7454/python/nnvm/frontend/coreml.py#L239)). Should I remove them in this PR?

tqchen: only concatenate support multiple input value

PR #187, opened 2017-10-20
Change the way to import ONNX model. (cf. #186)

imai-lm:
As I wrote in #186 , This seems a better way to import ONNX model.

Issue #186, opened 2017-10-19
onnx.load() outputs a model, not a graph... doesn't it?


imai-lm:
I guess the following part of code is wrong:
https://github.com/dmlc/nnvm/blob/69bea128fcd7d06a130160d66a6f2c33e4bfd5e9/tests/python/frontend/onnx/test_graph.py#L8-L9
This produces an error, like `AttributeError: 'ModelProto' object has no attribute 'initializer'`. 
I guess It should be:
```
 onnx_model = onnx.load(onnx_file) 
 onnx_sym, params = nnvm.frontend.from_onnx(onnx_model.graph) 
```
This produces another error, though. 

p.s. The version of ONNX is 0.2.

PENGUINLIONG:
That's right, you can PR it. Why these tests haven't been run so far...

BTW, What's the new error produced?

imai-lm:
> BTW, What's the new error produced?

When I just ran it as it is, an IndexError occured:
```
python nnvm/tests/python/frontend/onnx/test_graph.py
Traceback (most recent call last):
  File "nnvm/tests/python/frontend/onnx/test_graph.py", line 24, in <module>
    test_super_resolution_example()
  File "nnvm/tests/python/frontend/onnx/test_graph.py", line 21, in test_super_resolution_example
    compare_graph(fname, symbol, ishape=(1, 1, 224, 224))
  File "nnvm/tests/python/frontend/onnx/test_graph.py", line 9, in compare_graph
    onnx_sym, params = nnvm.frontend.from_onnx(onnx_model.graph)
  File "/root/.local/lib/python3.5/site-packages/nnvm-0.8.0-py3.5.egg/nnvm/frontend/onnx.py", line 397, in from_onnx
    sym, params = g.from_onnx(graph)
  File "/root/.local/lib/python3.5/site-packages/nnvm-0.8.0-py3.5.egg/nnvm/frontend/onnx.py", line 284, in from_onnx
    out = out[0]
IndexError: list index out of range
```
But this is because the loaded model (`onnx_file`) is bad (see #185). So I tried it with a model that is exported from PyTorch, TypeError occured like:
```
Traceback (most recent call last):
  File "nnvm-onnx-import.py", line 10, in <module>
    sym, params = nnvm.frontend.from_onnx(model.graph)
  File "/Users/imai/.local/lib/python3.5/site-packages/nnvm-0.8.0-py3.5.egg/nnvm/frontend/onnx.py", line 397, in from_onnx
[...]

Issue #185, opened 2017-10-19
model_zoo/super_resolution.onnx is not legal


imai-lm:
I guess [`tests/python/frontend/onnx/model_zoo/super_resolution.onnx`](https://github.com/dmlc/nnvm/blob/master/tests/python/frontend/onnx/model_zoo/super_resolution.onnx) should be a model in ONNX format, but it doesn't pass the current ONNX checker (ver 0.2).

Suppose the code `check.py` like the following:
```python
import onnx


print('ONNX version: ' + onnx.__version__)
model = onnx.load('super_resolution.onnx')
try:
    onnx.checker.check_model(model)
    print('valid model.')
except ValueError as e:
    print(e)
```
This produces an exception like this:
```
$ python check.py
ONNX version: 0.2
The model does not have an ir_version set properly.
```

tqchen:
close this as we moved discussion to #186

PR #189, opened 2017-10-20
Change the way to import ONNX model. (#186, #187)

imai-lm:
As I wrote in #186 , This seems a better way to import ONNX model.
(This is a reproduction of #187.)


tqchen:
please do a  rebase against the current master, the current commit appears to contain contents that reverts changes in coreml that is not part of onnx

python/nnvm/frontend/coreml.py
```diff
@@ -246,7 +246,9 @@ def ConcatLayerParams(op, insyms, symtab):
         insyms = [insyms]
     if op.sequenceConcat:
         raise NotImplementedError("Sequence Concat not supported")
-    ret = _sym.concatenate(*insyms, axis=1)
+    ret = insyms[0]
+    for i in range(1, len(insyms)):
```
tqchen: please do a rebase so this revert do not happen

imai-lm:
@tqchen I did it. I have failed to manage the commits, but now seems to succeed.

tests/python/frontend/onnx/test_graph.py
```diff
@@ -5,8 +5,8 @@
 from model_zoo import super_resolution
 
 def compare_graph(onnx_file, nnvm_sym, ishape):
-    onnx_graph = onnx.load(onnx_file)
-    onnx_sym, params = nnvm.frontend.from_onnx(onnx_graph)
+    onnx_model = onnx.load(onnx_file)
+    onnx_sym, params = nnvm.frontend.from_onnx(onnx_model.graph)
```
tqchen: I think a property would be check onnx_model.version and then change the code accordingly. So the current test pass and we can merge things in gradually

imai-lm: Neat Idea. I will try.

imai-lm: Well, what do you think about the version of `onnx`, not a model but the imported module? If you fix it to v0.2, you should update the `onnx` installed on the testing environment.

imai-lm: The older `onnx` (that is installed on the testing environment) will try to load a proto file as `GraphProto`, and v0.2 is going to load one as 'ModelProto'.

tqchen: The current test is built upon 0.1. I am just suggesting we do things like try and get ir version field of proto before we proceed so code works for both versions and the transition can be smooth. Otherwise we can update the env and the importer at the same time, but need to grantee test pass when merging

imai-lm: As far as reading [this issue on ONNX project](https://github.com/onnx/onnx/issues/119), versioning of an ONNX model is quite fragile at the moment. So an alternative is to check the version of ONNX module and change the code accordingly.

tqchen: Sound good to me

tqchen: can you start the change so that test can pass and we can merge the updates in?

imai-lm: yes, I'm working on it.

imai-lm: ah... the primary version of ONNX is devastating...

masahi:
@imai-lm does test_graph.py pass for you? It's not working for me with the latest onnx.

imai-lm:
@masahi It doesn't, but this is not because of the code, but because the used model is old and doesn't match the latest onnx library.  (see #185)

masahi:
@imai-lm alright, I hope nnvm team would support onnx 0.2 soon.

PR #191, opened 2017-10-22
Change as graph.input can be either ValueInfoProto or string (#186)

imai-lm:
Since ONNX v0.2, the type of `GrapProto.input` has changed from `string` to `ValueInfoProto`.


imai-lm:
I made a mistake. I'll withdraw it for now...

PR #196, opened 2017-10-24
[TVM] Update tvm and benchmark script

tqchen:


tqchen:
cc @Laurawly  please confirm the regression on K80

Laurawly:
@tqchen Both of the models converge on K40c.

Issue #194, opened 2017-10-24
How should Gemm operator in ONNX be converted?


imai-lm:
In #186, for coverting Gemm operator of ONNX, @zhreshold wrote:
> So basically Gemm == sym.dense(use_bias=False), adding it to the _convert_map should work.

But for me it seems difficult; NNVM's dense operator is `Y=XW^T+b`, where the 2nd operator `W` will be transposed at the calculation. However, ONNX's Gemm operator is actually `Y = alpha * A * B + beta * C`,  where the 2nd operator `B` should not be transposed (unless the attribute `transB` is non-zero).

Is there a better way to solve this? Or should I add another operator like `nnvm.symbol.dense_transpose` that calculates `Y=XW+b`, for example?

tqchen:
We can just insert a transpose op for the cast that is not covered by dense for now. Since it is not yet bottleneck, we can have internal optimizer to pattern match and optimize layer

tqchen:
For background, most efficient implementation of gemm actually would do transposition inside  before doing the dot, so there is likely no perf loss if we do it right.

imai-lm:
Well, the current implementation of `nnvm.frontend.onnx()` is based on [`_convert_map`](https://github.com/dmlc/nnvm/blob/master/python/nnvm/frontend/onnx.py#L102), where the conversion is defined with one-on-one basis (i.e. one operator is converted to one operator). Should we change this architecture a bit and modify it from upper levels?

zhreshold:
@imai-lm Yes, you can modify the one-to-one basis and take/return symbols directly just like this #175

imai-lm:
OK, I will try it anyway.

zhreshold:
@imai-lm How's your progress? I'm going to have 0.2 supported.

masahi:
@imai-lm @zhreshold I'm interested in this too. I need a solid onnx loading support to test our new tvm backend.

imai-lm:
@masahi Well, I have been coding it, but it doesn't work with my model, so far. I'm looking into it so please wait for a moment.

imai-lm:
I'm still working, but you can watch it if you are interested: https://github.com/dmlc/nnvm/compare/master...imai-lm:add-onnx-gemm-spike

masahi:
@imai-lm thanks. Is this gemm op used by gemm based conv ops?

imai-lm:
[...]

PR #202, opened 2017-10-27
[Frontend] ONNX frontend v0.2 support

zhreshold:
- [x] v0.2 support
- [x] gemm
- [x] some tests
- [ ] tutorial

@tqchen 
Test docker will need to be updated with torch installed. The pytorch installed with pip cannot support onnx and we should build pytorch locally or wait until the updated wheel to be released.

@imai-lm @masahi


masahi:
After this pr is merged, I am going to test it with various onnx models.

python/nnvm/frontend/onnx.py
```diff
@@ -95,10 +140,32 @@ def _batch_norm():
...
+        # get number of channels
+        channels = _infer_channels(inputs[1], nodes, params, renames)
+        # Y = alpha * A * B + beta * C
+        alpha = float(attr['alpha'])
+        beta = float(attr['beta'])
+        transA = int(attr['transA'])
+        transB = int(attr['transB'])
+        if transA:
+            inputs[0] = _sym.transpose(inputs[0], axes=(1, 0))
+        if transB:
```
imai-lm: If I understand it, this line should be `if not transB:`. Is it right?

imai-lm:
@zhreshold I have tried this branch with my pytorch-trained model, but failed (the same problem with what I mentioned in #194 occurred). And it seems that test models you added don't include `Gemm` layers, which are generated from `torch.nn.Linear`. Please try AlexNet, for example.

tqchen:
if you can verify locally, we can temporary remove the onnx unittests and renable until the docker setup is ready. 

Otherwise, we can also try directly download models from the onnx model website

tqchen:
Two things

- We should not keep putting onnx binary proto into the repo, this will make the source code repo explode.
- Maybe we can try to have script that downloads from onnx/models
- The regression test can check onnx versions, if mismatch, print a warning and quit, so we won't be caught by upstream sudden change

Pytorch support should be verified, but if it is a bit hard to put it into the docker, we can use the script downloading pre-packed model in onnx for now

imai-lm:
@tqchen I can do some tests locally, but I do not have enough platforms (e.g. I don't have machines with GPUs for now). So you should establish the testing environments on your side.

masahi:
I can confirm @imai-lm's finding. When I tried to load alexnet with this pr, I got

zhreshold:
@masahi No worry, this is fixed, after cleaning some test cases, I will send updated PR

masahi:
@zhreshold Great.

imai-lm:
LGTM. It causes another error in my environment, but it seems another issue, irreverent to this fix.

imai-lm:
At least in my environment, `tests/python/frontend/onnx/test_graph.py` causes an error like the following:
```
LM-MacBook-27:nnvm imai$ python tests/python/frontend/onnx/test_graph.py 
[14:03:45] /Users/imai/nnvm/dmlc-core/include/dmlc/./logging.h:308: [14:03:45] src/top/nn/convolution.cc:54: Operator conv2d(strides=(1, 1), channels=1, use_bias=False, kernel_size=(5, 5), groups=1, dilation=(1, 1), padding=(2, 2), name=conv2d4) expects weight's shape to be [1,1,5,5], but got [64,1,5,5].

(snip)

Traceback (most recent call last):
  File "tests/python/frontend/onnx/test_graph.py", line 24, in <module>
    test_super_resolution_example()
  File "tests/python/frontend/onnx/test_graph.py", line 21, in test_super_resolution_example
    compare_graph(fname, symbol, ishape=(1, 1, 224, 224))
  File "tests/python/frontend/onnx/test_graph.py", line 15, in compare_graph
    g1 = g1.apply("InferShape").apply("SimplifyInference")
  File "/Users/imai/.local/lib/python3.5/site-packages/nnvm-0.8.0-py3.5.egg/nnvm/graph.py", line 234, in apply
    check_call(_LIB.NNGraphApplyPasses(self.handle, npass, cpass, ctypes.byref(ghandle)))
  File "/Users/imai/.local/lib/python3.5/site-packages/nnvm-0.8.0-py3.5.egg/nnvm/_base.py", line 72, in check_call
    raise NNVMError(py_str(_LIB.NNGetLastError()))
nnvm._base.NNVMError: [14:03:45] src/top/nn/convolution.cc:54: Operator conv2d(strides=(1, 1), channels=1, use_bias=False, kernel_size=(5, 5), groups=1, dilation=(1, 1), padding=(2, 2), name=conv2d4) expects weight's shape to be [1,1,5,5], but got [64,1,5,5].
```

zhreshold:
@tqchen Passed after refreshing the old model.

tqchen:
Thanks! this is merged

imai-lm:
@zhreshold Thanks!

masahi:
@zhreshold Is Imagenet model loading supported?
The following script is not working.

```
import nnvm
import tvm
import nnvm.compiler
import onnx
from PIL import Image
import numpy as np
from torch.autograd import Variable
import torch.onnx
import torchvision

dummy_input = Variable(torch.randn(1, 3, 224, 224))
model = torchvision.models.alexnet(pretrained=True)
torch.onnx._export(model, dummy_input, "imagenet.proto", verbose=True)

onnx_file = "imagenet.proto"
onnx_graph = onnx.load(onnx_file)
sym, params = nnvm.frontend.from_onnx(onnx_graph)
sym = nnvm.sym.softmax(sym)

x = np.zeros((1, 3, 224, 224))
target = 'opencl'
shape_dict = {'data': x.shape}
graph, lib, params = nnvm.compiler.build(sym, target, shape_dict, params=params)

```

PR #217, opened 2017-11-03
[TOP] add conv2d_transpose

Huyuwei:


python/nnvm/top/nn.py
```diff
@@ -126,6 +126,38 @@ def schedule_conv2d(attrs, outs, target):
...
+    assert dilation == (1, 1), "not support dilate now"
+    assert groups == 1, "only support groups == 1 for now"
+    out = topi.nn.conv2d_transpose_nchw(inputs[0], inputs[1], strides, padding)
+    if attrs.get_bool("use_bias"):
+        bias = inputs[2]
+        bias = topi.expand_dims(bias, axis=1, num_newaxis=2)
+        out = topi.broadcast_add(out, bias)
+    output_padding = attrs.get_int_tuple("output_padding")
+    out = topi.nn.pad(out,
+        [0, 0, 0, 0], [0, 0, output_padding[0], output_padding[1]])
```
Huyuwei: or should this be out = topi.nn.pad(out, [0, 0, output_padding[0], output_padding[1]], [0, 0, 0, 0]) ?

tqchen: I think it should be in pad after

PR #218, opened 2017-11-03
Fix parameter loading for gluon Conv2dTranspose

masahi:
Somehow Conv2DTranspose in gluon stores its fliter as [in_channels, out_channels, ...].
However, conv2d_transpose in NNVM expects [out_channels, in_channels, ...]. 

```
In [7]: out_channels = 128

In [8]: conv = nn.Conv2D(out_channels, 4, 1, 0, in_channels = 64)

In [9]: conv_t = nn.Conv2DTranspose(out_channels, 4, 1, 0, in_channels = 64)

In [10]: conv.weight.shape
Out[10]: (128, 64, 4, 4)

In [11]: conv_t.weight.shape
Out[11]: (64, 128, 4, 4)
```

This causes an error when loading gluon models that use Conv2DTranspose.
For example, [this script](https://gist.github.com/masahi/50582fee8580d285791f694c1f61e93d) gives the following error.
```
$ python load_dcgan.py 
[09:46:33] /home/masa/tvm/dmlc-core/include/dmlc/logging.h:308: [09:46:33] src/top/nn/convolution.cc:161: 
Operator conv2d_transpose(use_bias=False, dilation=(1, 1), groups=1, padding=(0, 0), channels=512, kernel_size=[4, 4], output_padding=(0, 0), layout=NCHW, strides=(1, 1), name=conv2d_transpose0) 
expects weight's shape to be [512,100,4,4], but got [100,512,4,4].
``` 

This PR fixes this problem. Let me know if there is a better way to do it.

With this commit, I am able to load and run DCGAN with NNVM.


masahi:
cc @zhreshold

tqchen:
@zhreshold can you confirm if it is problem of the nnvm op or the frontend?

python/nnvm/frontend/mxnet.py
```diff
@@ -358,8 +358,13 @@ def from_mxnet(symbol, arg_params=None, aux_params=None):
         sym = symbol(data)
         sym = _from_mxnet_impl(sym, {})
         params = {}
-        for k, v in symbol.collect_params().items():
-            params[k] = tvm.nd.array(v.data().asnumpy())
+        for block in symbol._children:
```
zhreshold: This does not fix nested blocks for gluon models. Also, symbols are not fixed by this hack.
I think we should add a transpose op on top of weights inside Deconvolution Op.

tqchen:
The real problem seems to be we need to update our convention of the weight of the conv2d_tranpose in nnvm, as the reversed pattern is usually used(to match the symmetry of the backwar op). This is also in consistent with onnx and tensorflow

@zhreshold can you do a change to update this?

tqchen:
maybe @Huyuwei is a  better person doing so, since we also need to update topi's convention.

@Huyuwei can you do this? change weight convention of conv2d_tranpose to (in_channel, out_channel, h, w)

Huyuwei:
@tqchen ok, I will do it.

PR #219, opened 2017-11-03
Update deploy.md

WolffyChen:
Add an example in c++.


docs/how_to/deploy.md
```diff
@@ -35,3 +35,84 @@ out = tvm.nd.empty(shape)
...
+#include <algorithm>
+
+int main()
+{
+    // tvm module for compiled functions
+    tvm::runtime::Module mod_dylib = tvm::runtime::Module::LoadFromFile("deploy.so");
+
+    // json graph
+    std::ifstream json_in("deploy.json", std::ios::in);
+    std::stringstream json_ss;
```
PENGUINLIONG: `<sstream>` should be included for the use of `stringstream`. BTW, you can copy only once:

```cpp
std::string json_data(std::istream_iterator<char>(json_in), std::istream_iterator<char>());
```

docs/how_to/deploy.md
```diff
@@ -35,3 +35,84 @@ out = tvm.nd.empty(shape)
...
+    int dtype_code = kFloat;
+    int dtype_bits = 32;
+    int dtype_lanes = 1;
+    int device_type = kCPU;
+    int device_id = 0;
+
+    DLTensor* x;
+    int in_ndim = 4;
+    int64_t in_shape[4] = {1, 3, 224, 224};
+    TVMArrayAlloc(in_shape, in_ndim, dtype_code, dtype_bits, dtype_lanes, device_type, device_id, &x);
```
PENGUINLIONG: `TVMArrayFree()` should be called to free allocated resources.

docs/how_to/deploy.md
```diff
@@ -35,3 +35,84 @@ out = tvm.nd.empty(shape)
...
+    TVMArrayAlloc(out_shape, out_ndim, dtype_code, dtype_bits, dtype_lanes, device_type, device_id, &y);
+
+    // get the function from the module(get output data)
+    tvm::runtime::PackedFunc get_output = mod.GetFunction("get_output");
+    get_output(0, y);
+
+    // get the maximum position in output vector
+    std::vector<float> output(static_cast<float*>(y->data), static_cast<float*>(y->data) + 1000);
+    auto max_iter = std::max_element(output.begin(), output.end());
+    auto max_index = std::distance(output.begin(), max_iter);
```
PENGUINLIONG: Iterators are abstraction of pointer, so there is no need to place data collected into a container. Pointers can be passed directly.

docs/how_to/deploy.md
```diff
@@ -35,3 +35,84 @@ out = tvm.nd.empty(shape)
 get_output(0, out)
 print(out.asnumpy())
 ```
+
+An example in c++.
+```cpp
```
PENGUINLIONG: Maybe add an empty line?

docs/how_to/deploy.md
```diff
@@ -35,3 +35,84 @@ out = tvm.nd.empty(shape)
...
+    // get the function from the module(set input data)
+    tvm::runtime::PackedFunc set_input = mod.GetFunction("set_input");
+    set_input("data", x);
+
+    // get the function from the module(load patameters)
+    tvm::runtime::PackedFunc load_params = mod.GetFunction("load_params");
+    load_params(params_arr);
+
+    // get the function from the module(run it)
+    mod.GetFunction("run")();
```
tqchen: get function first before run, as run can be used multiple times

WolffyChen:
Thank you for your advice. @PENGUINLIONG @tqchen

docs/how_to/deploy.md
tqchen: Please use the iterator initializer method

docs/how_to/deploy.md
```diff
@@ -35,3 +35,86 @@ out = tvm.nd.empty(shape)
...
+An example in c++.
+```cpp
+#include <dlpack/dlpack.h>
+#include <tvm/runtime/module.h>
+#include <tvm/runtime/registry.h>
+#include <tvm/runtime/packed_func.h>
+
+#include <fstream>
+#include <sstream>
+#include <iterator>
```
PENGUINLIONG: Since you are not using `stringstream` it's reasonable to remove this.

pallabdatta:
Hi @tqchen, @WolffyChen 

I am trying to access the learned weights (params) with a graph pass on the C++ side. For..e.g. I see all the node params but I can't seem to have access to the learned weights..Note: within a pass, I haven't yet built the runtime module, i..e the (graph, lib, params = nnvm.compiler.build(...)) Can we save the params after the following:

onnx_graph = onnx.load('network.onnx')
sym, params = nnvm.frontend.from_onnx(onnx_graph) ? before we go into the compiler.build and loaded it directly via some C++ calls? 
Can you please help me with this?
thanks a lot in advance,
Pallab

tqchen:
@pallabdatta please open new issues for new questions. On your question, we can use nnvm.compiler.save_param_dict and load function in here https://github.com/dmlc/nnvm/blob/master/python/nnvm/compiler/param_dict.py

Note that they are PackedFunc and is also accessible in c++ side, see def in https://github.com/dmlc/nnvm/blob/master/src/compiler/graph_runtime.cc#L138 and

http://docs.tvmlang.org/dev/runtime.html#packedfunc

PR #223, opened 2017-11-06
Update deploy.md

WolffyChen:
Update dlpack device and type enumerations.


WolffyChen:
Oh! I added some commits that of my own repository since I am not so familiar with github. Should I withdraw it?

tqchen:
all the commits are squashed during merge, so it is fine. Thanks for the update

PR #260, opened 2017-11-28
Adde support for CoreML Permute layers and "same" padding in Pooling layers

notorca:
This PR depends on https://github.com/dmlc/tvm/pull/675


tqchen:
Let us consider another way of doing this without changing spec of the current core op. As this is redundant with padding option. Likely we can get away with supporting more padding and shape information. @zhreshold can you look into this?

tqchen:
The Perumate Layer support LGTM, please open a separate PR for that

titikid:
Hi all, i faced this issue abt "same padding" when i try to use tinyYolo.coreML model from here https://s3-us-west-2.amazonaws.com/coreml-models/TinyYOLO.mlmodel    in from_coreML scripts
Any suggestion?

PR #270, opened 2017-12-07
API call to get symbol output count

cjolivier01:
MXNet Symbol.__getitem__ using list_outputs() is too expensive, when it only cares about the output count in most cases. 

Dependency of: https://github.com/apache/incubator-mxnet/pull/8989

Showing up as a hot spot:

list_outputs_profiling


tqchen:
Seems there is some CI issues, could have things to do with old tvm version in your commit, consider update tvm?

cjolivier01:
seems up to date...

cjolivier01:
fixed

PR #276, opened 2017-12-18
[GRADIENT] Add backward operator to enable backward graph 

yuruofeifei:
@kevinthesun and I are currently working on supporting backward graph in NNVM.

Our current plan is to add backward to most commonly used operators in top folder.

This PR is trying to gather feedback from community, more operators support will come in the following PR.


kevinthesun:
With this patch, only elementwise_add and sqrt operator are covered with backward. We think it would be good to have a complete framework first and then add more operators. Same idea apply to potential  parameter update graph pass. Currently there is no optimizer operator defined in nnvm. We can start with naive SGD and add more latter.

include/nnvm/node.h
```diff
@@ -141,7 +141,8 @@ inline NodeEntry MakeNode(
     std::string node_name,
     std::vector<NodeEntry> inputs,
     std::unordered_map<std::string, std::string> attrs =
-    std::unordered_map<std::string, std::string>()) {
+    std::unordered_map<std::string, std::string>(),
+    NodePtr fwd_node = nullptr) {
```
kevinthesun: Add param documentation?

tqchen:
For most of the new simple operators like sqrt, we prefer not do add them as backward operator (with reference to forward version), but instead define FGradient by returning composition of new forward operators.

Bulk backward operator is only needed when there is need of shape inference like conv2d

yuruofeifei:
@tqchen @ZihengJiang @kevinthesun @piiswrong 

I have addressed Tianqi's comment and added tests for simple operators.

Please let me know your thoughts.

tests/python/compiler/test_top_level1.py
```diff
@@ -56,6 +127,19 @@ def test_log():
         y_np = np.log(data)
         np.testing.assert_allclose(out.asnumpy(), y_np, atol=1e-5, rtol=1e-5)
 
+        # backward
+        graph._set_symbol_list_attr('grad_ys', y)
```
ZihengJiang: Consider add a python function for this

kevinthesun: I'll work on this to add creating backward graph for both c_api and python api.

yuruofeifei:
Currently, nnvm does not have ones or zeros ops, added fill/fill_like operator to serve the same functionality. 
@kevinthesun and I think current ops are complete for testing backward graph generation flow after this PR. 
Can you please review? @tqchen 
Thanks!

tqchen:
please update docs/top.rst to include the operators. cc @ZihengJiang for doing a code review, i have already done a rough pass

yuruofeifei:
@ZihengJiang @yidawang @piiswrong

tqchen:
there is a fix of tutorial commit at most just now, please rebase

yuruofeifei:
Test passed

ZihengJiang:
LGTM

PR #279, opened 2017-12-23
Add create backward graph

kevinthesun:
1. Add c_api for create full forward-backward graph. 
2. Add some internal operators which are needed by creating backward graph.
3. Modify python graph create API to support backward graph.


tqchen:
Need to be very cautious about explicit introduction of concept of backward here. Instead, we can simply use the gradient pass. Please refer to implementation in tinyflow

tqchen:
Close this for now as the gradient pass requires quite different api

tqchen:
see https://github.com/tqchen/tinyflow/blob/master/python/tinyflow/_base.py#L68

kevinthesun:
OK. Will look a this and change accordingly.

PR #280, opened 2017-12-24
Add gradient graph

kevinthesun:
Add python function to create gradient symbol.
Add necessary operators for calling gradient pass.
@tqchen @ZihengJiang @yuruofeifei


python/nnvm/graph.py
```diff
@@ -13,6 +13,10 @@
 from ._base import GraphHandle, SymbolHandle
 from ._base import check_call
 from .symbol import Variable, Symbol, Group as _Group
+try:
+    from .symbol import ones_like
+except ImportError:
+    print("ones_like operator is required for creating gradient.")
```
tqchen: use warnings.warn

src/top/tensor/elemwise.cc
```diff
@@ -247,6 +248,25 @@ NNVM_REGISTER_INIT_OP(fill)
...
 .set_support_level(1);
 
+// Generate constant zero
+NNVM_REGISTER_INIT_OP(__zero__)
+.describe(R"code(Return constant value zero
+
+)code"  NNVM_ADD_FILELINE)
+.set_support_level(1);
+
+NNVM_REGISTER_INIT_OP(_zeros)
```
ZihengJiang: what's the difference between __zero__ and _zeros?

yuruofeifei: I think Yao is using __zero__ to construct a constant zero scalar.
I think we can use _zeros to do the same.

tqchen: one thing we need to make sure is to be consistent as numpy.

src/top/tensor/elemwise.cc
```diff
@@ -260,8 +273,38 @@ as the input array
...
+.describe(R"code(Return an array of ones with the same shape and type
+as the input array.
+
+)code")
+.add_argument("data", "Symbol", "The input")
+.set_attr<FGradient>(
+  "FGradient", [](const NodePtr& n,
+                  const std::vector<NodeEntry>& ograds){
+    return std::vector<NodeEntry>{
+      MakeNode("zeros_like", n->attrs.name + "_grad",
```
ZihengJiang: ones_like

src/top/tensor/elemwise.cc
```diff
@@ -260,8 +273,38 @@ as the input array
   "FGradient", [](const NodePtr& n,
                   const std::vector<NodeEntry>& ograds){
     return std::vector<NodeEntry>{
-      MakeNode("fill_like", n->attrs.name + "_zero",
-               {n->inputs[0]}, {{"value", "0"}})
+      MakeNode("zeros_like", n->attrs.name + "_grad",
```
ZihengJiang: full_like

PR #286, opened 2017-12-30
[SYMBOL] Enable Get Index from Symbol

ZihengJiang:


tqchen:
it does not make too much sense to get index from a grouped symbol, an error should be raised, @ZihengJiang can you followup a pr to fix this?

PR #291, opened 2018-01-02
Fix docstring of NNGraphGetJSONAttr

iblis17:
according to the implementation
https://github.com/dmlc/nnvm/blob/0aa3ff6fd561789b27242f257593962080f2de88/src/c_api/c_api_graph.cc#L63-L83


tqchen:
the return value is not part of success flag. The return value indicate whether there are exception thrown and returns 0 if nothing throws. This is done via NNVM_API_END macro

PR #292, opened 2018-01-04
dump lowered IR when debug logging

masahi:
This can be useful to check how lowered IR looks after fusion.

Also fixes a minor py2 issue.


python/nnvm/compiler/build_module.py
```diff
@@ -101,6 +101,7 @@ def _lower(sch, inputs, func_name, graph):
     try:
         f = tvm.lower(sch, inputs, name=func_name)
         logging.debug("lower function %s", func_name)
+        logging.debug(tvm.lower(sch, inputs, simple_mode=True))
```
tqchen: change to ```logging.debug("%s", tvm.lower(sch, inputs, simple_mode=True))```

then tvm.lower(sch, inputs, simple_mode=True) won't be called if debug is disabled

PR #298, opened 2018-01-10
Upsampling op support

masahi:
The follow up to dmlc/tvm#772. The test should pass after that PR is merged.

MXNet frontend is also included.


tqchen:
please update this pr

masahi:
done, I've updated tvm to the latest one

PR #300, opened 2018-01-11
[GRADIENT] Register more gradient operators

yuruofeifei:
This PR registered more common gradient operators in order to generate full forward backward graph used for training.
The test cases covered how to generate a gradient graph for simple CNN networks with correct shapes and types infer.

The next 2 or 3 PR will 
1. Design and implement optimizer interface in python.
2. Fill the gaps that some necessary operators (forward, backward) are not registered  in TVM. 
3. Test execution for training graph.

Please let me know your thoughts.
@ZihengJiang @piiswrong


docs/top.rst
```diff
@@ -54,6 +55,9 @@ This level enables fully connected multi-layer perceptron.
    nnvm.symbol.softmax
    nnvm.symbol.log_softmax
    nnvm.symbol.pad
+   nnvm.symbol.assign
+   nnvm.symbol.make_loss
```
tqchen: make loss seems to be not necessary for default gradient, i think we should avoid introduce this for now

include/nnvm/top/tensor.h
```diff
@@ -158,16 +158,35 @@ struct ReduceParam : public dmlc::Parameter<ReduceParam> {
...
+struct InitOpWithScalarParam : public dmlc::Parameter<InitOpWithScalarParam> {
+  TShape shape;
+  int dtype;
+  double scalar;
+
+  DMLC_DECLARE_PARAMETER(InitOpWithScalarParam) {
+    DMLC_DECLARE_FIELD(shape).set_default(TShape());
+    DMLC_DECLARE_DTYPE_FIELD(dtype).set_default(kFloat32)
+      .describe("Target data type.");
+    DMLC_DECLARE_FIELD(scalar).describe("Scalar value to fill");
```
tqchen: The argument name need to be consistent with numpy, which means it should be value

include/nnvm/tuple.h
```diff
@@ -83,6 +83,32 @@ class Tuple {
...
+   * \brief Assign content to tuple from iterator.
+   * \param begin the beginning of iterator
+   * \param end the end of the iterator
+   * \param d_first the beginning of the destination range
+   * \tparam RandomAccessIterator iterator type
+   */
+  template<typename RandomAccessIterator>
+  inline void assign(RandomAccessIterator begin,
+                     RandomAccessIterator end,
+                     ValueType* d_first) {
```
tqchen: This helper is bit abnormal, let us not include it in tuple as member, but put it as helper function somewhere.

src/top/nn/nn.cc
```diff
@@ -82,8 +118,23 @@ NNVM_REGISTER_ELEMWISE_UNARY_OP(relu)
...
 )code" NNVM_ADD_FILELINE)
+.set_attr<FGradient>(
+  "FGradient", [](const NodePtr& n,
+                  const std::vector<NodeEntry>& ograds) {
+    return MakeGradNode("_relu_grad", n,
+                        {ograds[0], n->inputs[0]});
+})
 .set_support_level(1);
 
+NNVM_REGISTER_OP(_relu_grad)
```
tqchen: not register gradient for simple operators like relu, instead simply return a composed graph containing the gradient

src/top/tensor/elemwise.cc
```diff
@@ -452,64 +448,65 @@ NNVM_REGISTER_ELEMWISE_BINARY_SCALAR(__rpow_scalar__)
...
+                  const std::vector<NodeEntry>& ograds){
+    CHECK_EQ(ograds.size(), 1);
+    std::vector<NodeEntry> ret;
+    for (size_t i = 0; i < n->inputs.size(); i++) {
+      ret.push_back(ograds[0]);
+    }
+    return ret;
+  });
+
+NNVM_REGISTER_ELEMWISE_UNARY_OP(make_loss)
```
tqchen: Do not use make_loss function, instead gradient attach ones_like by default

src/top/tensor/reduce.cc
```diff
@@ -122,6 +139,11 @@ Example::
 
 )code" NNVM_ADD_FILELINE);
 
+NNVM_REGISTER_REDUCE_OP_GRAD(_sum_grad)
```
tqchen: do not register separate gradient op for sum, use identity directly

docs/top.rst
```diff
@@ -54,6 +55,9 @@ This level enables fully connected multi-layer perceptron.
    nnvm.symbol.softmax
    nnvm.symbol.log_softmax
    nnvm.symbol.pad
+   nnvm.symbol.assign
```
tqchen: for now let us not include assign in the graph, we can use it to get the update and do assignments in the outside

src/top/tensor/transform.cc
```diff
@@ -58,8 +58,27 @@ Example::
...
 .add_argument("data", "Tensor", "Input data.")
+.set_attr<FGradient>(
+  "FGradient", [](const NodePtr& n,
+                  const std::vector<NodeEntry>& ograds){
+    return MakeGradNode("_flatten_grad", n,
+                        {ograds[0], n->inputs[0]});
+})
 .set_support_level(1);
 
+NNVM_REGISTER_OP(_flatten_grad)
```
tqchen: this can be reshape_like

tqchen:
A general message here is that we need to be cautious when introducing backward ops, as they are no longer differentiable. A better approach is to always map gradient to normal ops and rely on operator fusor to fuse the simple ops together

tqchen:
Another thing to bear in mind is that when we introduce an operator that have same functionality as numpy, it should be compatible to numpy

yuruofeifei:
Addressed all the comments, only max_pool and conv2d have bulk ops now.

include/nnvm/top/tensor.h
```diff
@@ -84,6 +89,14 @@ struct CastParam : public dmlc::Parameter<CastParam> {
   }
 };
 
+struct IndicatorParam : public dmlc::Parameter<IndicatorParam> {
```
tqchen: can we simply use greater and less as numpy did?

docs/top.rst
```diff
@@ -77,6 +81,8 @@ This level enables typical convnet models.
    :nosignatures:
 
    nnvm.symbol.reshape
+   nnvm.symbol.reshape_like
+   nnvm.symbol.expand_like
```
tqchen: is expand_like duplicating with reshape_like?

include/nnvm/top/tensor.h
```diff
@@ -158,16 +179,49 @@ struct ReduceParam : public dmlc::Parameter<ReduceParam> {
...
+
+struct ElementWiseReduceParam : public dmlc::Parameter<ElementWiseReduceParam> {
+  int num_args;
+  DMLC_DECLARE_PARAMETER(ElementWiseReduceParam) {
+    DMLC_DECLARE_FIELD(num_args).set_lower_bound(1)
+      .describe("Number of inputs to be reduced.");
+  }
+};
+
+struct DotParam : public dmlc::Parameter<DotParam> {
```
tqchen: the additional transpose operator is not consistent with numpy, if we want to keep it, we should use a different name, say matmul to be consistent with tensorflow.

yuruofeifei: sure, changed to matmul

include/nnvm/top/tensor.h
yuruofeifei: I think keeping indicator function is more useful here:
1. Unified interface in core src here is more concise as they share lots of properties including same gradient function. 
2. The operator can be potentially extended to generate mask array based on rules like equal to some value or max over certain axises. I also took a look at less and greater operator, it can be composed of broadcast sub and indicator function.

docs/top.rst
yuruofeifei: no, reshape_like is like reshape, the size of the both input and output is the same.
expland_like is more like composition of unsqueeze and expand dims, it does expand the tensor. It is used for construction of sum grad.

include/nnvm/top/tensor.h
tqchen: Thanks for digging deep this issue. Can you give an example of what do you mean by concise?

In general, we should avoid creating functions with user-facing names (like name indicator) when its application is not user-facing (depend on flags that user don't like to wrirte). For non-user facing functions, it is always a good idea to use underscore. This is a good way to avoid legacy issues later.

On the other hand, we should avoid creating underscore functions at all when possible and reuse existing user-facing functions, as likely high order gradient will be well defined.

These are general guidelines and I would like to discuss specific details in here

yuruofeifei: I agree with what you say, passing flags to operator is not user friendly. 
By concise, I mean "implementation-friendly" :)
I will change the code to align with numpy. Thank you for review.

yuruofeifei:
@tqchen all the comments are addressed, is there any other issues for this PR?

include/nnvm/top/tensor.h
tqchen: For greater, less. The axis and exclude parameter are not consistent with numpy. Should they be removed, or rename the function with an underscore.

docs/top.rst
```diff
@@ -28,6 +28,7 @@ This level enables fully connected multi-layer perceptron.
    :nosignatures:
 
    nnvm.symbol.dense
+   nnvm.symbol.dot
```
tqchen: dot->matmul

src/top/tensor/elemwise.cc
```diff
@@ -452,64 +449,78 @@ NNVM_REGISTER_ELEMWISE_BINARY_SCALAR(__rpow_scalar__)
...
+with 1.0 if (left < right), otherwise 0.0 element-wise.
+
+)code" NNVM_ADD_FILELINE)
+.add_argument("lhs", "Tensor", "First input")
+.add_argument("rhs", "Tensor", "Second input")
+.set_num_inputs(2)
+.set_attr<nnvm::FInferShape>("FInferShape", ElemwiseShape<2, 1>)
+.set_support_level(1);
+
+NNVM_REGISTER_INDICATOR_OP(max_mask)
```
tqchen: numpy non-compatible functions, consider use _max_mask?

yuruofeifei: Changed.

include/nnvm/top/tensor.h
yuruofeifei: For now, it's not used. I changed them to numpy compatible.

tqchen:
Thanks for doing the updates and act on reviews, this is now merged

PR #301, opened 2018-01-12
make shape inference of BatchNorm layout neutral

guopinglong:
The original shape inference logic is very specific to NCHW. This change is to make it a little be more layout neutral. So other layouts can be incorporated more easily.


guopinglong:
Nice comment. Refactored to use the axis variable directly instead of allocating a new one (layout).

src/top/nn/nn.cc
```diff
@@ -117,19 +117,33 @@ DMLC_REGISTER_PARAMETER(BatchNormParam);
...
       << "Input:[data, gamma, beta, moving_mean, moving_var]";
   CHECK_EQ(out_shape->size(), 3U);
   const TShape &dshape = in_shape->at(0);
   if (dshape.ndim() == 0) return false;
-  TShape bshape({dshape[1]});
-  NNVM_ASSIGN_INPUT_SHAPE(attrs, *in_shape, 1, bshape);
-  NNVM_ASSIGN_INPUT_SHAPE(attrs, *in_shape, 2, bshape);
-  NNVM_ASSIGN_INPUT_SHAPE(attrs, *in_shape, 3, bshape);
-  NNVM_ASSIGN_INPUT_SHAPE(attrs, *in_shape, 4, bshape);
+  auto bshape = [&](int axis) {
```
tqchen: do we really need switch in here? would dshape[axis] suffice?

guopinglong:
Yes. The change should be minor in this case.

guopinglong:
Test case added.

Issue #304, opened 2018-01-13
Keras layer UpSampling2D not supported


thefiddler:
I have successfully built and install tvm, topi and nnvm with the openblas and opencl backends.

While trying to convert a [simple keras model](https://github.com/dmlc/nnvm/files/1628389/simple.yml.txt), I am encountering the following error:
```bash
Traceback (most recent call last):
  File "compile_nnvm.py", line 6, in <module>
    sym, params = nnvm.frontend.from_keras(model)
  File "/Users/thefiddler/.virtualenvs/nnvm/lib/python3.6/site-packages/nnvm-0.8.0-py3.6.egg/nnvm/frontend/keras.py", line 445, in from_keras
    _check_unsupported_layers(model)
  File "/Users/thefiddler/.virtualenvs/nnvm/lib/python3.6/site-packages/nnvm-0.8.0-py3.6.egg/nnvm/frontend/keras.py", line 396, in _check_unsupported_layers
    raise ValueError("Keras layer {} not supported.".format(type(layer).__name__))
ValueError: Keras layer UpSampling2D not supported.
```

What would be the simplest way to add support for the UpSampling2D layer?

Edit: I am noticing that UpSampling2D is explicitly commented out in [keras.py:370](https://github.com/dmlc/nnvm/blob/master/python/nnvm/frontend/keras.py#L370).

Edit 2: I am preparing a PR to add support for keras upsampling.

thefiddler:
PR has been merged, closing!

PR #305, opened 2018-01-13
Add support for keras upsampling

thefiddler:
Map keras 'UpSampling*D' operators to topi upsampling. AFAICT, topi only supports symmetric upsampling, where all dimensions are equal - this PR will raise an error if this condition does not hold true. (Please correct me if I am wrong!)

Fixes issue #304

### DIFF
--- a/python/nnvm/frontend/keras.py
+++ b/python/nnvm/frontend/keras.py
@@ -274,21 +274,21 @@ def _convert_upsample(insym, keras_layer, _):
 
     if upsample_type == "UpSampling1D":
         h = keras_layer.size
-        params = {'scale': [h]}
+        params = {'scale': h}
     elif upsample_type == "UpSampling2D":
         h, w = keras_layer.size
         if h != w:
             raise TypeError("Unsupported upsampling type with different axes size : {}"
-                            .format(keras_layer))
+                            .format(keras_layer.size))
         params = {'scale': h}
     elif upsample_type == "UpSampling3D":
         h, w, d = keras_layer.size
         if h != w or w != d:
             raise TypeError("Unsupported upsampling type with different axes size : {}"
-                            .format(keras_layer))
+                            .format(keras_layer.size))
         params = {'scale': h}
     else:
-        raise TypeError("Unsupported upsampling type : {}".format(keras_layer))
+        raise TypeError("Unsupported upsampling type : {}".format(upsample_type))
 
     return _sym.upsampling(insym, **params)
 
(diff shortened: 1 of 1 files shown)


Huyuwei:
I think only UpSampling2D is supported by nnvm/tvm, and thus can be correctly converted.
https://github.com/dmlc/tvm/blob/master/topi/python/topi/nn/upsampling.py#L21
https://github.com/dmlc/nnvm/blob/master/src/top/nn/upsampling.cc#L19

thefiddler:
You are right, I have disabled the 1d and 3d variants. We can re-enable them once they are added to TVM.

python/nnvm/frontend/keras.py
```diff
@@ -268,6 +268,31 @@ def _convert_pooling(insym, keras_layer, symtab):
             raise TypeError("Unsupported pooling type : {}".format(keras_layer))
 
 
+def _convert_upsample(insym, keras_layer, _):
+    _check_data_format(keras_layer)
+    upsample_type = type(keras_layer).__name__
+
+    if upsample_type == "UpSampling1D":
+        h = keras_layer.size
+        params = {'scale': [h]}
```
Huyuwei: should remove [] ?

python/nnvm/frontend/keras.py
```diff
@@ -268,6 +268,31 @@ def _convert_pooling(insym, keras_layer, symtab):
...
+    elif upsample_type == "UpSampling2D":
+        h, w = keras_layer.size
+        if h != w:
+            raise TypeError("Unsupported upsampling type with different axes size : {}"
+                            .format(keras_layer))
```
Huyuwei: is it clearer to print keras_layer.size?

python/nnvm/frontend/keras.py
```diff
@@ -268,6 +268,31 @@ def _convert_pooling(insym, keras_layer, symtab):
...
+        params = {'scale': h}
+    elif upsample_type == "UpSampling3D":
+        h, w, d = keras_layer.size
+        if h != w or w != d:
+            raise TypeError("Unsupported upsampling type with different axes size : {}"
+                            .format(keras_layer))
+        params = {'scale': h}
+    else:
+        raise TypeError("Unsupported upsampling type : {}".format(keras_layer))
```
Huyuwei: should be format(upsample_type)?

python/nnvm/frontend/keras.py
```diff
@@ -367,7 +392,8 @@ def _default_skip(insym, keras_layer, _): # pylint: disable=unused-argument
     # 'Cropping1D'             : _convert_cropping,
     # 'Cropping2D'             : _convert_cropping,
     # 'UpSampling1D'           : _convert_upsample,
-    # 'UpSampling2D'           : _convert_upsample,
+    'UpSampling2D'           : _convert_upsample,
```
Huyuwei: move this line to after ZeroPadding2D? it looks better that way

thefiddler:
Thanks for the review, I've implemented the suggested changes in the latest commit.

PR #306, opened 2018-01-13
Avoid crash when keras linear activation does not have alpha and beta defined

thefiddler:
When converting a keras network, I encountered a crash where a `linear` activation did not have an alpha and beta attribute defined. This PR explicitly checks for the existence of those attributes to avoid the crash and allow the conversion to proceed.


tqchen:
assign to @Huyuwei

Huyuwei:
Thanks for the fix! @thefiddler

Issue #307, opened 2018-01-13
Incorrect padding for keras mode="same" in dilated convolution


thefiddler:
When converting keras models using `mode="same"` in convolutions and pooling, the [calculated padding shape](https://github.com/dmlc/nnvm/blob/master/python/nnvm/frontend/keras.py#L185-L197) is different between keras and nnvm. The attached program highlights the issue.

```bash
$ python3 test_padding.py
Output shape c1_same (None, 128, 128, 1)
Output shape c1_valid (None, 123, 122, 1)
Output shape c2_same (None, 128, 128, 1)
Output shape c2_valid (None, 118, 116, 1)
Output shape mp_same (None, 128, 128, 1)
Output shape mp_valid (None, 123, 122, 1)
Keras padding shape c1_same:  (5, 6, 0)
Keras padding shape c2_same:  (10, 12, 0)
Keras padding shape mp_same:  (5, 6, 0)
NNVM padding shape c1_same: (5, 6, 0)
NNVM padding shape c2_same: (5, 6, 0)
NNVM padding shape mp_same: (5, 6, 0)
```

Note that you would need to add a print statement to python/nnvm/frontend/keras.py on lines 150 and 258 (or use a debugger) to observe the calculated values.

test_padding.py:
```python
import nnvm
from keras.models import Model
from keras.layers import *
from keras.layers.pooling import MaxPooling2D
from keras.optimizers import *
from keras.activations import linear

input = Input((128, 128, 1))
c1_same  = Convolution2D(1, (6, 7), dilation_rate=(1, 1), padding="same")(input)
c1_valid  = Convolution2D(1, (6, 7), dilation_rate=(1, 1), padding="valid")(input)
c2_same = Convolution2D(1, (6, 7), dilation_rate=(2, 2), padding="same")(input)
c2_valid = Convolution2D(1, (6, 7), dilation_rate=(2, 2), padding="valid")(input)
mp_same  = MaxPooling2D((6, 7), strides=(1, 1), padding="same")(input)
mp_valid  = MaxPooling2D((6, 7), strides=(1, 1), padding="valid")(input)
model = Model([input], [c1_same, c1_valid, c2_same, c2_valid, mp_same, mp_valid])

print("Output shape c1_same", model.output_shape[0])
print("Output shape c1_valid", model.output_shape[1])
print("Output shape c2_same", model.output_shape[2])
print("Output shape c2_valid", model.output_shape[3])
print("Output shape mp_same", model.output_shape[4])
[...]

PR #308, opened 2018-01-13
(WIP) Match padding calculations to keras

thefiddler:
Update padding calculations to match keras. Keras bases its padding calculations on the kernel size and dilation (if any), not on the shape of the input tensor. This fixes the padding mode="same" for dilated convolutions (issue #307).

By following the same approach, we can now also convert models with undefined input shapes, i.e. `(3, None, None)` rather than `(3, 224, 224)`.

### DIFF
--- a/python/nnvm/frontend/keras.py
+++ b/python/nnvm/frontend/keras.py
@@ -2,6 +2,8 @@
 """Keras frontend."""
 from __future__ import absolute_import as _abs
 import sys
+from enum import Enum
+from collections import Sequence
 import numpy as np
 import tvm
 from .. import symbol as _sym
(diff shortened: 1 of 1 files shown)


Huyuwei:
As far as I know, padding calculations in keras do depend on the input size, at least when using tensorflow backend https://www.tensorflow.org/versions/r0.12/api_docs/python/nn/convolution

Now 'same' padding for dilated convolutions has a bug as reported, and this should be fixed by
```python
kernel_h = (kernel_h - 1) * dilation[0] + 1
kernel_w = (kernel_w - 1) * dilation[1] + 1
```

Even if we convert it correctly on the frontend, now only dilation==1 is supported by TVM on the backend.

thefiddler:
According to the [v1.4 tensorflow documentation](https://www.tensorflow.org/api_docs/python/tf/nn/convolution), the padding only depends on the kernel size and dilation, not the input shape:
If padding == "SAME":
```python
output_spatial_shape[i] = ceil(input_spatial_shape[i] / strides[i])
```

If padding == "VALID":
```python
output_spatial_shape[i] = ceil((input_spatial_shape[i] - (spatial_filter_shape[i]-1) * dilation_rate[i]) / strides[i]).
```
(Of course the output shape depends on the input shape, but the _padding_ only depends on the kernel and dilation rate.)

The reason this matters is that using the input shape in the padding calculation means that we cannot convert networks with unknown input shapes, since the input shape is not known during the conversion process. If we use only the kernel & dilation, then we can convert those networks correctly.

This is important for us, as we are evaluating using NNVM/TVM in production. :)

Huyuwei:
@thefiddler Thanks for pointing out. I will investigate it later. Could you please confirm that after the modification, the tests pass https://github.com/dmlc/nnvm/blob/master/tests/python/frontend/keras/test_forward.py

thefiddler:
There are issues, investigating.

Huyuwei:
@thefiddler I still hold the view that padding depends on input size, as documented in https://www.tensorflow.org/versions/r0.12/api_docs/python/nn/convolution.

Assume input = [32, 32], kernel=[3, 3], stride=[2, 2], according to your method, pad_top=1, pad_bottom=1; according to my method, pad_top=0, pad_bottom=1, a test will show the second is right.

After #322, pooling with same pad should pass tests with all input sizes, pool sizes, strides. Dilated conv should also be padded correctly, even though it's not supported by TVM backend.

thefiddler:
After spending quite a bit more time, it seems that padding indeed does rely on the input size. I think.the difference is that tensorflow calculates padding dynamically at runtime, so it can deal with tensors of any shape. I am closing this in favour of #322

PR #311, opened 2018-01-16
Add more Keras tests

thefiddler:
This PR increases the number of tests for the Keras frontend and makes the convolution, pooling and upsampling tests exhaustive (i.e. it tests a wider range of different input size, kernel and stride parameters). It also adds a test for forward shape inference, which does not currently work.

The exhaustive tests reveal some consistent failure cases in the convolution and pooling implementation. In particular, when using the LLVM backend, I get a consistent failure for Conv2D and SeparableConv2D when using input size 15x15, with kernel size 1x1 and stride 3x3. Similar failures can also be observed in the implementation of MaxPooling2D.

Open questions:
1. Is the goal to match the Keras/Tensorflow output in 100% of the cases (shape * kernel * stride) or is it ok if only the most common layouts are supported?
2. Is it allowed to have print statements inside the testing code? It is quite hard to identify the failing case otherwise.

This PR (or a similar PR) should be merged before #308

### DIFF
--- a/tests/python/frontend/keras/test_forward.py
+++ b/tests/python/frontend/keras/test_forward.py
@@ -128,11 +128,11 @@ def test_forward_conv_odd_small():
 
 def test_forward_conv_even_strided():
     print("test_forward_conv_even_strided")
-    data = keras.layers.Input(shape=(32,32,4))
-    x = keras.layers.Conv2D(filters=10, kernel_size=(3,3), strides=(2, 1), padding='same')(data)
-    x = keras.layers.Conv2D(filters=10, kernel_size=(4,4), strides=(1, 2), padding='same')(x)
-    x = keras.layers.Conv2D(filters=10, kernel_size=(3,3), strides=(1, 2), padding='valid')(x)
-    x = keras.layers.Conv2D(filters=10, kernel_size=(4,4), strides=(2, 1), padding='valid')(x)
+    data = keras.layers.Input(shape=(64,64,4))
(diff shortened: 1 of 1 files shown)


thefiddler:
There is a crash happening here that appears similar to #312 

Any help would be appreciated.

thefiddler:
What is the best way to proceed here? The new tests reveal a number of failing cases, should we disable those until the right fixes are merged? Or should we relax the tests to not demand 100% identical results with keras/tensorflow in all cases?

The crash errors should probably be fixed in either case.

Huyuwei:
I can reproduce test_forward_pooling error locally, and I will submit a PR to fix it soon.

test_forward_conv passes for all inputs on my machine. Does it fail only with llvm and pass with cuda on your machine? @thefiddler 

As for the conv_transpose error, I think it's related to the TVM backend.

thefiddler:
@Huyuwei I have only tested on llvm so far, let me try on cuda and report back.

Huyuwei:
@thefiddler  A better way to push this forward is to submit an issue or PR for each error, since some of them are due to the keras converter and some due to TVM backend. 
It's not practical to add so many tests, it makes the CI build time too long (2min -> 5min). I expect the test to cover most basic cases. Other errors or crashes can be addressed through issue and PR.

thefiddler:
Ok, I will try to split this in parts and reduce the exhaustive search to something more reasonable that still covers the failure cases I'm seeing.

PR #313, opened 2018-01-16
[TUTORIAL] add tutorial for mali gpu

merrymercy:


tqchen:
need to update the tvm to latest version

tqchen:
to avoid the cl deadlock during rpc server fork, always use use_popen=True as in https://github.com/dmlc/tvm/blob/master/tutorials/deployment/cross_compilation_and_rpc.py#L104 for local RPC server

PR #314, opened 2018-01-17
Update frontend for keras 2.1.3 compatibility

thefiddler:
Keras keeps obsessively renaming properties every second release. Keras 2.1.3 renames `layer.inbound_nodes` to `layer._inbound_nodes`.

This PR fixes the crashes that occur when converting a Keras 2.1.3 model to NNVM. It is backwards compatible with previous Keras versions so it should not break anything else.

### DIFF
--- a/python/nnvm/frontend/keras.py
+++ b/python/nnvm/frontend/keras.py
@@ -484,9 +484,14 @@ def from_keras(model):
             symtab.get_var(keras_layer.name, must_contain=False)
         else:
             predecessors = []
-            print(keras_layer.name)
             inbound_nodes = keras_layer.inbound_nodes if hasattr(keras_layer, 'inbound_nodes') \
-                        else keras_layer._inbound_nodes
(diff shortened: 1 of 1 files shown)


python/nnvm/frontend/keras.py
```diff
@@ -349,7 +349,14 @@ def _convert_concat(insym, keras_layer, _):
 
 
 def _convert_reshape(insym, keras_layer, _):
-    return _sym.reshape(insym, keras_layer.shape)
+    shape = keras_layer.shape if hasattr(keras_layer, 'shape') else \
```
Huyuwei: is it clearer to write the code as:
if hasattr(keras_layer, 'shape'):
   ...
elif hasattr(keras_layer, 'target_shape'):
   ...
else:
   raise TypeError("message")

Huyuwei: Ok, just ignore the last comment

PR #315, opened 2018-01-17
BinaryBroadcastShape Inference

guopinglong:
For BatchNorm ops of an inference graph, SimplifyInference (simplify_inference.cc) will replace them with broadcast_mul/broadcast_add stuff. Therefore, the layout information should pass along this graph rewrite pass to ensure correct shape inference.
This is only the initial tentative fix to this issue. Comments are greatly appreciated.


guopinglong:
Looks like this one is messed up with previous changes. Will close this one and resubmit.

PR #321, opened 2018-01-17
Fix BinaryBroadcastShape() when layout is not NCHW

guopinglong:
For BatchNorm ops of an inference graph, SimplifyInference (simplify_inference.cc) will replace them with broadcast_mul/broadcast_add stuff. Therefore, the layout information should pass along this graph rewrite pass to ensure correct shape inference of BinaryBroadcastShape.
This is only the initial tentative fix to this issue. Comments are greatly appreciated.


guopinglong:
I don't know why preliminary checks have failed. 
I can not reproduce them locally.

guopinglong:
Check results include this error message. While this PR is definitely related to binarybroadcast op, everything works fine in my local environment. So, I tend to conclude that this error is caused by other reasons.
ERROR: test_infer_shape.test_broadcast_binary

----------------------------------------------------------------------

Traceback (most recent call last):

  File "/usr/local/lib/python2.7/dist-packages/nose/case.py", line 197, in runTest

    self.test(*self.arg)

  File "/workspace/tests/python/unittest/test_infer_shape.py", line 254, in test_broadcast_binary

    check((4, 1), (4), (4, 4))

  File "/workspace/tests/python/unittest/test_infer_shape.py", line 251, in check

    sdict = infer_shape(z)

  File "/workspace/tests/python/unittest/test_infer_shape.py", line 8, in infer_shape

    g = g.apply("InferShape")

  File "/workspace/python/nnvm/graph.py", line 235, in apply

    check_call(_LIB.NNGraphApplyPasses(self.handle, npass, cpass, ctypes.byref(ghandle)))

  File "/workspace/python/nnvm/_base.py", line 72, in check_call

    raise NNVMError(py_str(_LIB.NNGetLastError()))

NNVMError: Error in operator y: [03:37:16] /workspace/dmlc-core/include/dmlc/././any.h:286: Check failed: type_ != nullptr The any container is empty requested=N4nnvm3top20BinaryBroadcastParamE

guopinglong:
Thanks for your comment. My patch works find in my local tvm/nnvm with no problem.
What I am confused is the error message popped up when I submit this PR. After this, I will consider your your point on switching to other (potentially) simpler implementation alternatives.

guopinglong:
Thanks --> @yuruofeifei 
Check still failed. Why this can not be reproduced locally?

guopinglong:
Don't know why the check still failed, yet could not be reproduced locally.

tqchen:
Not sure why test failed. However, we cannot change the spec of binary broadcast add operator to diverge from numpy. So adding layout here is not desirable. Can we comeup with a solution without changing the op def?

I am going to close this for now and let us open a new PR once we have the solution

PR #322, opened 2018-01-17
[Keras] fix convert_pooling with same pad

Huyuwei:


thefiddler:
It might make sense here to add a default values for `in_h` and `in_w` when they are `None`, possibly with a warning, to allow conversion of variable size networks (i.e. with `Input((None, None, C))`) to proceed. The default value could be a power of two (e.g. 128) since those tend to be pretty common in practice.

In that case, a converted network will work correctly as long as the user ensures the input size can be evenly divided by all pooling / strided-convolution layers in the network. In the ideal scenario, the padding would be calculated at runtime by TVM, but I don't think that is possible right now. The second-best case would allow more networks to be successfully converted and executed.

I can provide a PR if required.

Issue #325, opened 2018-01-18
Keras frontend: incorrect results when adding more than one tensor together


thefiddler:
The following code gives incorrect results:
```python
from keras.layers import *
from keras.model import Model

r = []
input = Input((128, 128, 1))
x = Conv2D(8, (3, 3), padding="same")(input)
r.append(x)
x = Conv2D(8, (3, 3), padding="same")(x)
r.append(x)
x = Conv2D(8, (3, 3), padding="same")(x)
x = add([x, *r])

model = Model(input, x)

verify_keras_frontend(model)
```

If I change `x = add([x, *r])` to:
```python
x = add([x, r[0]])
x = add([x, r[1]])
```
it works correctly.

tqchen:
close by #326 thanks to @thefiddler

PR #326, opened 2018-01-18
Fix keras frontend elementwise-ops for lists with len>2  (fixes issue #325)

thefiddler:
This PR adds a test for element-wise addition. Currently, the first test passes correctly and the second one fails.

See also #325


thefiddler:
The issue was that the keras frontend was only considering the first two symbols in the list. I have pushed a simple fix to this PR.

PR #333, opened 2018-01-24
[DEBUG] Add BuildConfiguration option to disable GraphFusePartition.

srkreddy1238:
Build configuration option to disable fusing of multiple nodes into one operation.
Helpful while developing & debugging frontends.


tqchen:
this is disabled automatically if opt_level is set to 0

Issue #336, opened 2018-01-25
Unable to build ONNX model using NNVM


Mani-Sadhasivam:
Hello,

I have a Caffe2 model exported to ONNX format using the below instructions:
https://github.com/onnx/tutorials/blob/master/tutorials/Caffe2OnnxExport.ipynb

When I try to build this model using NNVM I'm getting the below error:
https://pastebin.ubuntu.com/26459575/

Below commands are used to build the model:
```
onnx_graph = onnx.load('sqeezenet.onnx')
sym, params = nnvm.frontend.from_onnx(onnx_graph)
```
The same commands works well for the `super_resolution` model

PENGUINLIONG:
It seems your NNVM is outdated. Could you try the latest version (cloned from the master branch) and see if it helps?

Mani-Sadhasivam:
Rebasing to latest commit doesn't solve the issue :(

Error Log: https://pastebin.ubuntu.com/26481727/

srkreddy1238:
Can you print and see check the operator name in [onnx.py](https://github.com/dmlc/nnvm/blob/c3cfe0ae4e9c2f2d4c4004666fd17b3c11d1efb8/python/nnvm/frontend/onnx.py#L487) ??

Siva

Mani-Sadhasivam:
Operator name: Conv

srkreddy1238:
Looks like some extra attributes from mxnet for Conv which is not addressed from frontend.

Try printing [attrs](https://github.com/dmlc/nnvm/blob/c3cfe0ae4e9c2f2d4c4004666fd17b3c11d1efb8/python/nnvm/frontend/onnx.py#L87) for super resolution and your model.


Siva

srkreddy1238:
@Mani-Sadhasivam 

Try https://github.com/dmlc/nnvm/pull/345

Siva

Mani-Sadhasivam:
@srkreddy1238 No luck :(

sqeezenet params: `{u'strides': (2L, 2L), u'pads': (0L, 0L, 0L, 0L), u'kernel_shape': (3L, 3L)}`

super_resolution params: `{u'dilations': (1L, 1L), u'strides': (1L, 1L), u'pads': (2L, 2L, 2L, 2L), u'group': 1L, u'kernel_shape': (5L, 5L)}`

srkreddy1238:
Pls share more details on this error.

I see the graph is generated as below.

```
Graph(%input_0,
      %param_0,
      %param_1,
[...]

PR #345, opened 2018-01-30
[Convolution] Error while importing onnx model with weights.

srkreddy1238:
Fix the use_bias based on the input parameters to use bias (Ex: sqeezenet). Ref #336


python/nnvm/frontend/onnx.py
```diff
@@ -95,7 +95,7 @@ def _impl(inputs, attr, params):
                 'dilations': ('dilation', (0, 0)),
                 'pads': ('padding', (0, 0), _revert_caffe2_pad),
                 'group': ('groups', 1)},
-            extras={'use_bias': False},
+            extras={'use_bias': True if len(inputs) == 3 else False},
```
tqchen: ```True if len(inputs) == 3 else False   --> len(inputs) == 3```

PR #334, opened 2018-01-24
Add optimizer

kevinthesun:
1. Add optimizer API.
2. Add SGD and Adam optimizers.
3. Add learning rate scheduler API.

Currently the state variables of optimizer, such as momentum and number of updates, need to be explicitly initialized when building the library. it might be able to improve this process.

PR #349, opened 2018-01-31
[TOP] Adding compute for more operators

yuruofeifei:
Adding compute for more operators to enable training.
All operators forward and backward compute results are compared with numpy.

List of operator added:
clip, matmul, elemwise_sum, full, full_like, ones, ones_like, zero, zeros_like, greater, less, expand_like, reshape_like, assign, block_grad

Some operators are dependent on TVM change: https://github.com/dmlc/tvm/pull/849
I will update this PR when that one got merged.

With this PR, the operator coverage is sufficient for simple training task like handwriting recognition, I will write an example.
 
@ZihengJiang @mli


tqchen:
@yuruofeifei close this and move to #334

Issue #343, opened 2018-01-28
[LANG] C++ Compiler API


tqchen:

srkreddy1238:
Thanks @tqchen  for consolidating everything at one place. 
Good direction towards native compilation support.

I could extend my support if needed here.

Siva

alex-weaver:
Hi @srkreddy1238, I'm looking at the first point at the moment (registering FTVMCompute ops via C++), which hopefully won't take too long. The second point though is the real crux of getting this to work - how do we handle finding the right schedule for an op, and possibly replacing the op with a target-dependent one (eg dense_cuda), and how should this fit into the nnvm architecture? Support with that aspect would be appreciated.

srkreddy1238:
@alex-weaver, I will try to put my head into this over this weekend. 

Thanks,
Siva

alex-weaver:
Having thought a bit about environment targeting, I think it is prudent to break it down into two issues.
1. How do we return the correct schedule from FTVMSchedule given the target string passed in?
2. Sometimes, an op should be replaced with a target-specific alternative (eg. dense_cuda), how and where should this happen?

## Proposal for issue 1
We give each schedule type a name, eg `injective`  `dense` `softmax` under which one generic and any number of device-specific schedules are registered.
There is a class called something like `ScheduleFactory` where there is one instance for each schedule type, stored in a `dmlc::Registry`.
ScheduleFactory has methods with signatures similar to the following
```
using FTVMScheduleBuilder = std::function<
  Schedule(const tvm::Target& target, const tvm::Array<tvm::Tensor>& outs)>;

ScheduleFactory& set_generic_schedule(const FTVMScheduleBuilder& value);
ScheduleFactory& add_schedule(const std::vector<std::string>& tags, const FTVMScheduleBuilder& value);
Schedule get_schedule(const tvm::Target& target, const tvm::Array<tvm::Tensor>& outs);
```
Where `set_generic_schedule` and `add_schedule` are used to register schedules, and ops that use this mechanism can then call `get_schedule` from within `FTVMSchedule` after parsing the target string using `tvm::Target::create`

## Issue 2
This appears harder to address. Since replacing the op is closely tied to a specific schedule, it would make some sense for `FTVMSchedule` above to be able to replace the op it is scheduling (question: does the call to FTVMSchedule occur too late in the graph compilation process for op replacement to work?)
This could then be used to extend ScheduleFactory with an extra method:
```
[...]

PR #359, opened 2018-02-06
Changed TOPI ops to use C++ schedules

alex-weaver:
Added a ScheduleFactory class to dispatch target-specific schedules, and TOPI ops now use this for scheduling rather than the python schedules.
#343


tqchen:
Let us spend more time to think about this before getting to a c++ api. ScheduleFactory is fine for c++ alone, but we have additional requirements for being able to interacting with the python side and interact with the existing generic dispatching module.

alex-weaver:
Ok that makes sense. If the python and C++ need to cooperate, then does this mean that the C++ should maintain the registry and should expose a way for python to register specializations?

clouds56:
Is there any update or anything I could here?

alex-weaver:
@clouds56 This PR is currently mostly blocked on the TVM PR https://github.com/dmlc/tvm/pull/892
once that one is resolved, I think there should only be one further obstacle to this PR working: in the current implementation of https://github.com/dmlc/tvm/pull/892 the C++ and Python schedules are completely separate, although the capability certainly exists for them to interoperate. Figuring out the details of that will probably be the main obstacle for this PR. Having some way for the C++ schedules to be registered by default, and then allow the python ones to override them is the main aim with that.

alex-weaver:
@tqchen Ok this is now updated to use the new C++ GenericFunc system in TVM for schedules

include/nnvm/top/util.h
```diff
@@ -0,0 +1,30 @@
...
+#include <tvm/build_module.h>
+#include <nnvm/base.h>
+#include <nnvm/compiler/op_attr_types.h>
+
+#include <string>
+
+namespace nnvm {
+namespace top {
+
+inline nnvm::compiler::FTVMSchedule MakeScheduleQuery(const std::string& name) {
```
tqchen: document what this is about.

PR #362, opened 2018-02-08
Flatten node injection before softmax as softmax expects the input in…

srkreddy1238:
… 2D (#336)


tqchen:
cc @zhreshold can you review this?

zhreshold:
`axis` is supported by softmax, therefore no need to flatten it.

masahi:
@srkreddy1238 I have a PR at dmlc/tvm#882 which includes support for arbitrary dimension softmax.

srkreddy1238:
Thanks for the axis support. It works good with graph compilation.

PR #364, opened 2018-02-09
Move dense compute back to python

kevinthesun:
Move dense compute registration back to python so that we can utilize newly added schedule in TVM.

@yidawang


yidawang:
This is for facilitating the GEMM scheduling work. We will probably move it to the C++ API once the scheduling is stable.

PR #369, opened 2018-02-13
WebGL end-to-end example

phisiart:
Local OpenGL execution succeeds.

The remote module can be successfully loaded. But crashes during execution
throwing an uncaught exception. The crash happens in the host LLVM module.
The browser dumps this stack trace. Need to figure out the problem.

```
___cxa_throw @ libtvm_web_runtime.js:1
via          @ libtvm_web_runtime.js:2948
Mqa          @ libtvm_web_runtime.js:2128
RN           @ libtvm_web_runtime.js:5017
_TVMFuncCall @ VM61:1                      <- This is the LLVM host module.
na           @ VM61:21
ftCall_iiii  @ libtvm_web_runtime.js:1
IHa          @ libtvm_web_runtime.js:1963
s_           @ libtvm_web_runtime.js:2484
Eha          @ libtvm_web_runtime.js:3946
cra          @ libtvm_web_runtime.js:5457
kJa          @ libtvm_web_runtime.js:2018
via          @ libtvm_web_runtime.js:2948
Mqa          @ libtvm_web_runtime.js:2128
n6           @ libtvm_web_runtime.js:3176
NAa          @ libtvm_web_runtime.js:5654
Ewa          @ libtvm_web_runtime.js:5653
FVa          @ libtvm_web_runtime.js:3043
lza          @ libtvm_web_runtime.js:5504
Asa          @ libtvm_web_runtime.js:5503
vGa          @ libtvm_web_runtime.js:5281
Nma          @ libtvm_web_runtime.js:4027
Oia          @ libtvm_web_runtime.js:4026
kJa          @ libtvm_web_runtime.js:2018
via          @ libtvm_web_runtime.js:2948
Mqa          @ libtvm_web_runtime.js:2128
Gma          @ libtvm_web_runtime.js:4786
eN           @ libtvm_web_runtime.js:4785
NI           @ libtvm_web_runtime.js:4803
wN           @ libtvm_web_runtime.js:2116
AK           @ libtvm_web_runtime.js:2112
WQ           @ libtvm_web_runtime.js:2967
u3           @ libtvm_web_runtime.js:2966
x4           @ libtvm_web_runtime.js:3060
Mla          @ libtvm_web_runtime.js:5228
kJa          @ libtvm_web_runtime.js:2018
via          @ libtvm_web_runtime.js:2948
Mqa          @ libtvm_web_runtime.js:2128
RN           @ libtvm_web_runtime.js:5017
ret          @ tvm_runtime.js:445
on_message   @ tvm_runtime.js:921
```


tutorials/define_and_compile_model_webgl.py
```diff
@@ -0,0 +1,206 @@
...
+# Parameters are initialized with Xavier initializer.
+# NNVM also supports other model formats such as MXNet, CoreML and ONNX.
+#
+# In this tutorial, we assume we will do inference on our device
+# and the batch size is set to be 1. Input images are RGB color
+# images of size 16 * 16. We can call the :any:`nnvm.symbol.debug_str`
+# to show the network structure.
+
+batch_size = 1
+num_class = 2
```
phisiart: Note: I use a smaller setting here. When using the original model, the js module crashes complaining "memory too small".

tutorials/define_and_compile_model_webgl.py
```diff
@@ -0,0 +1,206 @@
...
+module.set_input(**rparams)
+
+# set input data
+print('Setting input data')
+input_data = np.random.uniform(size=data_shape)
+module.set_input('data', tvm.nd.array(input_data.astype('float32')))
+
+# run
+print("module.run()")
+module.run()
```
phisiart: Note: the crash happens here.

phisiart:
Update: progress on debugging.

The crash happens when calling `__tvm_set_device` in host code. The body (a `std::function`) of the `PackedFunc` is empty, throwing a `std::bad_function_call`.

tqchen:
they should have been in here https://github.com/dmlc/tvm/blob/master/src/runtime/c_runtime_api.cc#L480

phisiart:
Update: progress on debugging.

When retrieving the `PackedFunc` `__tvm_set_device`, there seems to be an if statement:
```
if (...) {
  func = ...; // this looks like loading from some sort of cache
} else {
  func = TVMBackendGetFuncFromEnv(func_name, ...);
}
```

I manually changed the generated js module file so that it always goes through `TVMBackendGetFuncFromEnv`. Then the crash goes away.

phisiart:
Update: progress on debugging.

After changing the logic in the js module file, then we crash on the next `PackedFunc`. Let me temporarily change the logic of `CodeGenCPU` to only go through `TVMBackendGetFuncFromEnv` and see whether we can go to the end. If so, then I'll try to figure out the reason.

phisiart:
Update: progress on debugging.

It works.

phisiart:
@tqchen Update:

I tried to change [this line](https://github.com/dmlc/tvm/blob/master/src/codegen/llvm/codegen_cpu.cc#L486) from `LinkOnceAnyLinkage` to `PrivateLinkage` and the crash went away. This might be a possible fix?

tqchen:
that is a possible fix, seems to be problem of fastcomp though. Link once any allows us to only have one copy of handle if there are multiple objects get linked

tqchen:
any thoughts about "memory too small" complains, is there anyway we can fix this?

phisiart:
We can set a larger memory size or turn on auto memory growing in emcc. Let me try later. AFK now.

phisiart:
Update: I changed the TVM Makefile to set `-s TOTAL_MEMORY=1073741824` flag (total memory = 1GB instead of default 16MB). Now the original model runs successfully.

phisiart:
I also tried enabling automatic memory growth, but emcc says this feature is "not supported for shared module".

tqchen:
nice, related issue here kripken/emscripten#5589

tqchen:
Need to update tvm submodule and make sure CI pass. Because the tutorial is built from scratch in CI, we might need to skip the build optionally, when rpc is not available, or gl is not enabled and explains this issue. And put most of things in comment

tqchen:
We need to add a section on how to deploy final modules into the browser without RPC (some general tips are in here https://github.com/dmlc/tvm/tree/master/web#use-tvm-to-generate-javascript-library)

- Build things as usual
- If we don't want dynamic loading, save the library as system module and link it together with TVM runtime 
- Save the model as binary(possibly as base64)
- Provide a simple graph runtime wrapper(should in tvm's js runtime) as in python, which takes in graph_json, model_binary
    - The lib module is just ```tvm.systemLib```

- Alternatively, we can also enable dynamic loading which follows similar flow of RPC

phisiart:
Sorry for the delay. Please expect updates starting this weekend.

phisiart:
Update: working on this. I'm dealing with some weird exceptions.

phisiart:
Update: I've successfully built a WebGL TVM module that runs without RPC. This involves several tricky things. Let me first create a tutorial in TVM before coming back to NNVM.

phisiart:
Update: I've succeeded on creating a WebGL version of [get_started](https://github.com/dmlc/nnvm/blob/master/tutorials/get_started.py). Now I'll cleanup the code into a JS version of graph_runtime.

tqchen:
Actually, I think for WebGL, doing something like http://nnvm.tvmlang.org/tutorials/from_keras.html#sphx-glr-tutorials-from-keras-py would be more fun

tqchen:
See possible useful resources for preprocessing https://github.com/dmlc/mxnet.js

phisiart:
Update: The resnet18 in tutorials/define_and_compile_model succeeded. Let me first clean up this PR.

phisiart:
Update: The resnet18 model in [from_mxnet.py](https://github.com/dmlc/nnvm/blob/master/tutorials/from_mxnet.py) succeeded in local OpenGL and gives the same prediction result. I'll try WebGL tomorrow and change this PR to use that model.

tutorials/define_and_compile_model_webgl.py
```diff
@@ -0,0 +1,289 @@
...
+# for this graph on the target hardware, and the parameter blobs of
+# the model. During the compilation, NNVM does the graph-level
+# optimization while TVM does the tensor-level optimization, resulting
+# in an optimized runtime module for model serving.
+#
+# To generate the module library, TVM will first transfer graph IR into lower
+# intrinsic IR for the specified target backend, which is OpenGL in this
+# example. Then target backend will generate module library.
+
+if not tvm.module.enabled("opengl"):
```
tqchen: a bit too many opencl checks, maybe we can just use exit(0) here

tqchen: add a comment in the block, saying that because this tutorial is built in CI server, so the result may not be displayed, but user can download the notebook and run it

tutorials/define_and_compile_model_webgl.py
```diff
@@ -0,0 +1,289 @@
...
+# ------------------------
+# Now we can load the module back.
+
+import numpy as np
+from tvm.contrib import graph_runtime
+
+if not tvm.module.enabled("opengl"):
+    print("OpenGL backend not enabled. This tutorial cannot be run.")
+else:
+    loaded_lib = tvm.module.load(path_lib)
```
tqchen: Isolate deploy_local, deploy_rpc, deploy_web into three functions, and add proper comments

tutorials/resnet.html
```diff
@@ -0,0 +1,77 @@
+<html>
+
+<head>
+  <meta charset="UTF-8">
+  <title>TVM RPC Test Page</title>
```
tqchen: is it possible to move it to tutorials/web?

phisiart:
Update:

tutorials/define_and_compile_model_webgl.py
phisiart: Changed the 3 demos into separated functions.

tutorials/resnet.html
phisiart: Moved.

tutorials/define_and_compile_model_webgl.py
phisiart: Now only check once.

tqchen:
looks good @phisiart let me know if you need further changes, or if the current version can be merged

phisiart:
ready to be merged

tutorials/web/resnet.html
```diff
@@ -0,0 +1,103 @@
+<html>
+
+<head>
+  <meta charset="UTF-8">
+  <title>NNVM WebGL Test Page</title>
+</head>
+
+<body>
+  <h1>NNVM WebGL Test Page</h1>
+  <img src="data.png">
```
tqchen: Currently pre-processing is done in python side. Can we update the script to use javascript image processing instead? This is likely what most people want to know. reference in https://github.com/dmlc/mxnet.js/blob/master/classify_image.js#L33

PR #372, opened 2018-02-13
[NHWC] InferShape Layout conversion fix.

srkreddy1238:


tqchen:
just want to confirm the NHWC weight convention and what is the intended behavior. @yuruofeifei @yidawang @yzhliu can you confirm?

tqchen:
cc @piiswrong This seems to be a common convention of NHWC used in packages.

masahi:
@tqchen I don't know if it is common in other frameworks, but this was the format I figured out while I was trying to vectorize NHWC convolution.

Suppose the filter layout is [num_filter, height, width, in_channel]. Then I would vector-load from input channel dimension. But since input channel axis is one of the reduce axes,  I cannot do data parallel multiply-add-accumulate. Instead, I need to reduce over vector elements.

yzhliu:
@tqchen I'm implementing a more general layout system, in new design, weight has its own layout, e.g., OIHW (oc, ic, height, width). We can talk about it later.
For the changes here, honestly I don't know whether it is a common weight convention for NHWC either.

PR #376, opened 2018-02-19
Initial NHWC layout support

masahi:
Support NHWC layout for conv, pooling, upsampling, and softmax.
A simple test case is included. 

@srkreddy1238


tqchen:
nice work, can you add a testcase for NHWC case you added?

masahi:
what do you mean? A test case is already included at tests/python/compiler/test_nhwc_layout.py.

tqchen:
ah, sorry that i missed the code when i glance over, this can now be merged.

srkreddy1238:
I have a full model on NHWC with conv, pool & softmax.

I will test today and conclude the functionality :)

Siva

Issue #367, opened 2018-02-12
Importing from mxnet fails if variable is used twice


aseyboldt:
Converting an mxnet graph fails if a variable appears more than once in the graph:

```python
import nnvm
import mxnet as mx

a = mx.sym.var('a')
c = a + a
a_ = mx.nd.ones((10,))
nnvm.frontend.from_mxnet(c, {'a': a_})
```

This is on master 535d63e8faf4dfb339d91df686ccaecd3552311c

<details>

```pytb
ValueError                                Traceback (most recent call last)
<ipython-input-4-e468d259f583> in <module>()
      5 c = a + a
      6 a_ = mx.nd.ones((10,))
----> 7 nnvm.frontend.from_mxnet(c, {'a': a_})

~/git/nnvm/python/nnvm/frontend/mxnet.py in from_mxnet(symbol, arg_params, aux_params)
    352 
    353     if isinstance(symbol, mx.sym.Symbol):
--> 354         sym = _from_mxnet_impl(symbol, {})
    355         params = {}
    356         arg_params = arg_params if arg_params else {}

~/git/nnvm/python/nnvm/frontend/mxnet.py in _from_mxnet_impl(symbol, graph)
    315     if childs:
    316         op_name = symbol.attr('op_name')
--> 317         childs = [_from_mxnet_impl(c, graph) for c in childs]
    318         childs = [x for y in childs for x in _as_list(y)]  # expand group symbol
    319         node = _convert_symbol(op_name, childs, attr)

~/git/nnvm/python/nnvm/frontend/mxnet.py in <listcomp>(.0)
    315     if childs:
    316         op_name = symbol.attr('op_name')
--> 317         childs = [_from_mxnet_impl(c, graph) for c in childs]
    318         childs = [x for y in childs for x in _as_list(y)]  # expand group symbol
    319         node = _convert_symbol(op_name, childs, attr)

~/anaconda3/lib/python3.6/site-packages/mxnet/symbol/symbol.py in <genexpr>(.0)
     87         <Symbol _plus0>
     88         """
---> 89         return (self[i] for i in self.list_outputs())
     90 
     91     def __add__(self, other):

~/anaconda3/lib/python3.6/site-packages/mxnet/symbol/symbol.py in __getitem__(self, index)
    504                 if name == index:
    505                     if idx is not None:
--> 506                         raise ValueError('There are multiple outputs with name \"%s\"' % index)
    507                     idx = i
    508             if idx is None:

[...]

PR #383, opened 2018-02-25
[FRONTEND] fix same var used in single op

zhreshold:
Fix if same var in used in single op #367


tqchen:
see CI error https://github.com/dmlc/nnvm/pull/383

zhreshold:
__len__ was added 3 month ago, revert to len(list_outputs()) or update mxnet in CI?

tqchen:
I see, let us make things backward compatible first. As the CI env is a bit fragile atm

Issue #351, opened 2018-01-31
Build failure on Windows because of old submodule


AtuNuka:
Hi,

NNVM/TVM is failed to build in newest version on Windows in below environment.

NNVM Commit: dd4f6a9f95a45068252b2f32f7e07d1b07596740
OS: Windows 10 Pro 64bit
Compiler: Visual Studio 15 2017

```
2>C:\Users\nuka\nnvm\tvm\src\runtime\workspace_pool.cc(31): error C2059: syntax error: '.'
2>C:\Users\nuka\nnvm\tvm\src\runtime\workspace_pool.cc(31): error C2143: syntax error: missing ';' before '}'
2>C:\Users\nuka\nnvm\tvm\src\runtime\workspace_pool.cc(32): error C2059: syntax error: 'if'
2>C:\Users\nuka\nnvm\tvm\src\runtime\workspace_pool.cc(32): error C2143: syntax error: missing ';' before '{'
2>C:\Users\nuka\nnvm\tvm\src\runtime\workspace_pool.cc(32): error C2447: '{': missing function header (old-style formal list?)
2>C:\Users\nuka\nnvm\tvm\src\runtime\workspace_pool.cc(41): error C2059: syntax error: 'else'
2>C:\Users\nuka\nnvm\tvm\src\runtime\workspace_pool.cc(41): error C2143: syntax error: missing ';' before '{'
2>C:\Users\nuka\nnvm\tvm\src\runtime\workspace_pool.cc(41): error C2447: '{': missing function header (old-style formal list?)

<snip>
```

This is because ```tvm/src/runtime/workspace_pool.cc``` uses GNU struct initialization. And it seems that this has already been fixed in tvm project.

https://github.com/dmlc/tvm/commit/68b9aa9a05b5914a7c5cf2fce9fda13f4ea6f503#diff-06c69bd0cdfc13afd730c7658460da97

So, the real problem is that NNVM is pointing to the old TVM commit.
I'd appreciate to update a submodule TVM to the latest or proper version.

Thanks.

PENGUINLIONG:
Before submodule is updated, you can enter the `tvm` directory, `git pull` and rebuild, to use the latest version.

AtuNuka:
Thanks.
I will update manually for now.

I think it may be great if this operation is added to the Installation Guide.
Possibly, there may be people who have same issues.

If it is a general way, I'll send pull requests which adds this operations to NNVM Installation Guide.
But I'm not sure about it.

[...]

PR #384, opened 2018-03-06
[DOCS] add a solution to installation error because of the old submodule

AtuNuka:
Some user may fail to build because of the old submodule in future. (as I mentioned in #351 )
This patch adds one of the solution to the document.

In some case, this solution can not solve the installation error completely, but this document helps the new user solved it.

Related Issue: #351


docs/how_to/install.md
```diff
@@ -75,3 +76,14 @@ There are several ways to install the package, in all these cases the TVM librar
...
     cd python; python setup.py install --user; cd ..
     ```
+
+## Solution to Installation Error
+
+If you encounter the problem while installation process, you can solve by updating submodules to the latest commit set.
+To update submodules to the latest commit set, type the following command.
+
+```bash
+git submodule foreach git pull origin master
```
tqchen: change to git submodule update --init --recursive

AtuNuka: Fixed it. please check cb2eaf3

PR #391, opened 2018-03-13
[FIX] several bugs found when using NNVM

FinalTheory:
Fixed several bugs when using `nnvm` in my project, the diffs and corresponding details are listed below.

Environment: CentOS 6.3 x64, gcc 4.8.3, latest nnvm 770f6ff3ebea4e60775ce24ff41c8f95e7ed12f6

diff in `node.h`:

When using `MakeNode` with no additional attributes supplied, the `attrs.any` would be empty, which will cause a runtime error.

Consider following code:

```python
import nnvm
x = nnvm.sym.Variable('x')
w = nnvm.sym.Variable('w')
y = nnvm.sym.dense(x, w, units=128)
g = nnvm.compiler.graph_util.get_gradient_graph(y, x)
nnvm.compiler.graph_util.infer_shape(g, x=[100, 100])
```

Results in:

```
nnvm/dmlc-core/include/dmlc/././any.h:286: Check failed: type_ != nullptr The any container is empty requested=N4nnvm3top11MatMulParamE
```

diff in `plan_memory.cc`:

If a node ignores all its inputs, it means the remaining refcount `storage_ref_count[sid_in] == 1` is actually a reference used by another node, and this memory should not be recycled. So here we should add a checking to ensure that current node will consume at least 1 input, which could be safely released.

diff in `nn.cc`:

- gradient w.r.t input of relu should multiplies the gradient back-propagated from its output.
- a typo which makes the `list_input_names('aux_state')` to be empty, consider following code:

```python
import nnvm
a = nnvm.sym.Variable('a')
b = nnvm.sym.batch_norm(a)
b.list_input_names('aux_state')
```

diff in `matrix_op.cc`:

Consider following code:

```python
import nnvm
a = nnvm.sym.Variable('a')
b = nnvm.sym.Variable('b')
c = nnvm.sym.matmul(a, b)
g = nnvm.graph.create(c)
nnvm.compiler.graph_util.infer_shape(g, a=[10, 10], b=[10, 10])
```

The output is `([[10, 10], [10, 10]], [[10, 1, 10]])`, but the result of `infer_shape` should be [10, 10] instead of [10, 1, 10].

### DIFF
--- a/include/nnvm/node.h
+++ b/include/nnvm/node.h
@@ -146,11 +146,9 @@ inline NodeEntry MakeNode(
   NodePtr p = Node::Create();
   p->attrs.op = nnvm::Op::Get(op_name);
   p->attrs.name = std::move(node_name);
-  if (attrs.size() != 0) {
-    p->attrs.dict = attrs;
-    if (p->attrs.op->attr_parser) {
-      p->attrs.op->attr_parser(&(p->attrs));
-    }
+  p->attrs.dict = attrs;
+  if (p->attrs.op->attr_parser) {
+    p->attrs.op->attr_parser(&(p->attrs));
   }
   p->inputs = std::move(inputs);
   return NodeEntry{p, 0, 0};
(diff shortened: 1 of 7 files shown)


tqchen:
Thanks for the great catches, can you add test cases to the tests/python to cover the last two cases you mentioned?

tqchen:
thanks, this is merged

PR #394, opened 2018-03-15
[OP] PReLU Support

PariksheetPinjari909:
Depends on https://github.com/dmlc/tvm/pull/1008

### DIFF
--- a/include/nnvm/top/nn.h
+++ b/include/nnvm/top/nn.h
@@ -101,6 +101,17 @@ struct LeakyReLUParam : public dmlc::Parameter<LeakyReLUParam> {
   }
 };
 
+struct PReLUParam : public dmlc::Parameter<PReLUParam> {
+  double slope;
+  int axis;
+  DMLC_DECLARE_PARAMETER(PReLUParam) {
+    DMLC_DECLARE_FIELD(slope).set_lower_bound(0.0).set_default(0.0)
+        .describe("slope coefficient for the negative half axis."
+                  " Set this for non-channelwise slope");
+    DMLC_DECLARE_FIELD(axis).set_default(1)
+      .describe("Specify which shape axis the channel is specified.");
+  }
+};
 
 struct PadParam : public dmlc::Parameter<PadParam> {
   float pad_value;
--- a/include/nnvm/top/nn.h
+++ b/include/nnvm/top/nn.h
@@ -102,12 +102,8 @@ struct LeakyReLUParam : public dmlc::Parameter<LeakyReLUParam> {
 };
 
 struct PReLUParam : public dmlc::Parameter<PReLUParam> {
-  double slope;
   int axis;
   DMLC_DECLARE_PARAMETER(PReLUParam) {
-    DMLC_DECLARE_FIELD(slope).set_lower_bound(0.0).set_default(0.0)
-        .describe("slope coefficient for the negative half axis."
-                  " Set this for non-channelwise slope");
     DMLC_DECLARE_FIELD(axis).set_default(1)
       .describe("Specify which shape axis the channel is specified.");
   }
--- a/src/compiler/fold_scale_axis.cc
+++ b/src/compiler/fold_scale_axis.cc
@@ -347,10 +347,16 @@ NNVM_REGISTER_OP(relu)
 NNVM_REGISTER_OP(leaky_relu)
 .set_attr<FScaleAxisBackward>("FScaleAxisBackward", ReluScaleAxisBackward);
 
+NNVM_REGISTER_OP(prelu)
+.set_attr<FScaleAxisBackward>("FScaleAxisBackward", ReluScaleAxisBackward);
+
 NNVM_REGISTER_OP(relu)
 .set_attr<FScaleAxisForward>("FScaleAxisForward", ReluScaleAxisForward);
 
 NNVM_REGISTER_OP(leaky_relu)
+.set_attr<FScaleAxisForward>("FScaleAxisForward", ReluScaleAxisForward);
+
+NNVM_REGISTER_OP(prelu)
 .set_attr<FScaleAxisForward>("FScaleAxisForward", ReluScaleAxisForward);
 
 // property registration.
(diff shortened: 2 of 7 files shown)


include/nnvm/top/nn.h
```diff
@@ -101,6 +101,17 @@ struct LeakyReLUParam : public dmlc::Parameter<LeakyReLUParam> {
   }
 };
 
+struct PReLUParam : public dmlc::Parameter<PReLUParam> {
+  double slope;
```
tqchen: slope is not needed in PReLU

src/top/nn/nn.cc
```diff
@@ -419,6 +419,78 @@ NNVM_REGISTER_OP(leaky_relu)
 })
 .set_support_level(1);
 
+// prelu
+DMLC_REGISTER_PARAMETER(PReLUParam);
+
+inline uint32_t PReluNumInputs(const NodeAttrs& attrs) {
+  const PReLUParam& param = get<PReLUParam>(attrs.parsed);
```
tqchen: just make it take two inputs instead of making a complicated choice

tqchen: if there is no slope, it is simply relu

tests/python/compiler/test_top_level1.py
```diff
@@ -57,6 +57,76 @@ def forward(x):
...
+    a = sym.Variable("a")
+    y = sym.prelu(data=x, cslope=a)
+
+    def forward(x, a):
+        out = np.zeros(x.shape)
+        for b in range (x.shape[0]):
+            for c in range (x.shape[1]):
+                for h in range (x.shape[2]):
+                    for w in range (x.shape[3]):
+                        if x[b][c][h][w] < 0:
```
tqchen: use on-liner numpy style as in c++

src/top/nn/nn.cc
```diff
@@ -419,6 +419,56 @@ NNVM_REGISTER_OP(leaky_relu)
...
+
+NNVM_REGISTER_OP(prelu)
+.describe(R"code(Parametric version of a Rectified Linear Unit.
+It accepts two arguments: an input ``x`` and a channelwise slope ``W``
+and computes the output as :math:`PReLU(x) y = x > 0 ? x : W * x`,
+where :math:`*` is an elementwise multiplication for each sample in the
+
+)code" NNVM_ADD_FILELINE)
+.add_argument("data", "Tensor", "Input data.")
+.add_argument("cslope", "Tensor", "Input channelwise slope.")
```
tqchen: to make it consistent with leaky_relu, let us use alpha for this argument

src/top/nn/nn.cc
```diff
@@ -419,6 +419,56 @@ NNVM_REGISTER_OP(leaky_relu)
...
+
+  TShape oshape(dshape);
+  NNVM_ASSIGN_OUTPUT_SHAPE(attrs, *out_shape, 0, oshape);
+  return true;
+}
+
+NNVM_REGISTER_OP(prelu)
+.describe(R"code(Parametric version of a Rectified Linear Unit.
+It accepts two arguments: an input ``x`` and a channelwise slope ``W``
+and computes the output as :math:`PReLU(x) y = x > 0 ? x : W * x`,
```
tqchen: update comments according to the definition, change W to alpha

src/top/nn/nn.cc
```diff
@@ -419,6 +419,56 @@ NNVM_REGISTER_OP(leaky_relu)
...
+    return std::vector<std::string>{"data", "cslope"};
+  })
+.set_attr<FTVMCompute>(
+  "FTVMCompute", [](const NodeAttrs& attrs,
+                    const Array<Tensor>& inputs,
+                    const Array<Tensor>& out_info) {
+    const PReLUParam& param = nnvm::get<PReLUParam>(attrs.parsed);
+    return Array<Tensor>{ topi::prelu<float>(inputs[0], inputs[1], param.axis)};
+  })
+.set_support_level(1);
```
tqchen: support level->4

src/compiler/fold_scale_axis.cc
```diff
@@ -347,12 +347,18 @@ NNVM_REGISTER_OP(relu)
 NNVM_REGISTER_OP(leaky_relu)
 .set_attr<FScaleAxisBackward>("FScaleAxisBackward", ReluScaleAxisBackward);
 
+NNVM_REGISTER_OP(prelu)
+.set_attr<FScaleAxisBackward>("FScaleAxisBackward", ReluScaleAxisBackward);
```
tqchen: PReLU don't have the scale axis property, because the channel wise weight. please remove it for now from fold_axis optimization

src/top/nn/nn.cc
```diff
@@ -417,6 +417,56 @@ NNVM_REGISTER_OP(leaky_relu)
...
+                            std::vector<TShape> *in_shape,
+                            std::vector<TShape> *out_shape) {
+  const PReLUParam &param = nnvm::get<PReLUParam>(attrs.parsed);
+  TShape dshape = in_shape->at(0);
+  NNVM_ASSIGN_INPUT_SHAPE(attrs, *in_shape, 0, dshape);
+
+  // The case of parametric relu
+  CHECK_EQ(dshape.ndim(), 4) << "Input data should be 4D, but got " << dshape.ndim();
+  CHECK(size_t(param.axis) < dshape.Size()) <<
+        "Wrong axis ("  << param.axis << ")value. ";
```
tqchen: need to support shape inference so that slope shape can be deduced by the input shape

PariksheetPinjari909:
@tqchen the review comments are fixed. please check this also

tests/python/unittest/test_top_level3.py
```diff
@@ -16,8 +16,15 @@ def test_leaky_relu():
     y = sym.leaky_relu(x, alpha=0.1)
     assert(y.list_input_names() == ["x"])
 
+def test_prelu():
```
tqchen: add a test case that automatically infer the shape of alpha for given  data shape

PariksheetPinjari909: infershape is also tested as part of the test_prelu_nchw and test_prelu_nhwc in compiler/test_top_level1.py. Any shape mismatch, will throw error as part of this testcase. Please check whether this is sufficient or i need to add another testcase.

PariksheetPinjari909: @tqchen still need to add testcase for this? please check whether  test_prelu_nchw and test_prelu_nhwc in compiler/test_top_level1.py is fine? since it already foes the shape mismatch validation for data n alpha

PariksheetPinjari909:
@tqchen , updated the testcase for infershape. please check. Thank you for the review.

Issue #373, opened 2018-02-15
[matmul] Some queries


srkreddy1238:
referring to matrix_op i checked the documentation @ [matrix_op.cc](https://github.com/dmlc/nnvm/blob/535d63e8faf4dfb339d91df686ccaecd3552311c/src/top/tensor/matrix_op.cc#L53)

1: I saw that in source documentation and still unclear. A 2D multiplication goes wrong as explained below.

Out shape with input shapes [1, 2048] and [2048, 1008] resulted in [1, 1, 1008] .
Is this expected ? 

2: My other concern of compute & schedule not found which is raising error while building the graph.

Any help appreciated.

Siva

tqchen:
@kevinthesun can you followup on this issue?

kevinthesun:
There are some issues in matmul. shape is not correct. Will fix this together in https://github.com/dmlc/nnvm/pull/334

tqchen:
@kevinthesun fix there is a bug need to be fixed, we can just submit a separate PR

kevinthesun:
Will do that.

PR #396, opened 2018-03-19
fix matmul

jiajiechen:
@kevinthesun, I think the shape problem mentioned in https://github.com/dmlc/nnvm/issues/373 is already fixed, so I just added the `FTVMCompute`.

### DIFF
--- a/tests/python/compiler/test_top_level1.py
+++ b/tests/python/compiler/test_top_level1.py
@@ -8,45 +8,47 @@ import nnvm.compiler
 from nnvm.testing.config import ctx_list
 
 
-def helper(symbol, inputs, dtype,
-           np_forward, np_backward=None, **kwargs):
-    ishapes = {}
-    input_syms = []
+def helper(symbol, inputs, dtype, np_forward, np_backward=None, **kwargs):
+    input_shapes = {}
+    input_symbols = []
     np_inputs = {}
     for (k, v) in inputs.items():
-        ishapes.update({k: v[0]})
+        input_shapes.update({k: v[0]})
         np_inputs.update({k: np.random.uniform(size=v[0]).astype(dtype)})
         if len(v) > 1:
-            input_syms.append(v[1])
+            input_symbols.append(v[1])
 
     # put additional kwargs for np forward
     np_inputs_with_addtional_args = copy.copy(np_inputs)
     np_inputs_with_addtional_args.update(**kwargs)
 
     for target, ctx in ctx_list():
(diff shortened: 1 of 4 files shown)


tqchen:
@kevinthesun can you review?

kevinthesun:
Currently matmul on;y supports 2D inputs, it would be better if we can support N-D operation, similar to numpy.

tqchen:
It is a reasonable proposal, but if this PR is a bugfix, then we should merge this in for now. @kevinthesun can you confirm that?

tests/python/compiler/test_top_level1.py
```diff
@@ -308,6 +313,40 @@ def forward(x):
...
+    rhs = sym.Variable('rhs')
+    result = sym.matmul(lhs, rhs, transpose_a=False, transpose_b=True)
+
+    dtype = "float32"
+    inputs = {
+        'lhs': ((1, 3), lhs),
+        'rhs': ((4, 3), rhs),
+    }
+
+    helper(result, inputs, dtype, forward, transpose_b=True)
```
kevinthesun: Add test for backend?

tqchen:
error in testcase http://mode-gpu.cs.washington.edu:8080/blue/organizations/jenkins/dmlc%2Fnnvm/detail/PR-396/8/pipeline/

tqchen:
need rebase against the master

tqchen:
@jiajiechen please rebase against the master, let us merge this in after the rebase

tqchen:
close because of stable issue

wangshangsam:
Hi,

Is this issue fixed? As of today pulling from dmlc/tvm/nnvm, I'm still having the "Attribute FTVMCompute has not been registered for Operator matmul" issue, and this change doesn't seem to be merged in.

Thanks!

srkreddy1238:
not fixed, we don't have compute for matmul now. 
I had same issue and I used transpose+dense (with out bias) for 2D.

I will PR matmul compute with 2D support using dense. This can be used until we have topi support for higher order multiply.

wangshangsam:
Hi Siva,

Thanks for your response! The background of me having this issue is that, when you try to get the gradient graph for dense, it has matmul. Although I could somehow hack into the gradient graph and replace matmul with transpose + dense, this is not what I'm supposed to do from the perspective of the user of the tool. I would say that at current point even a 2D matmul is far more helpful than none since it's needed for gradients of a bunch of other operators.

Thanks!
Shang

PR #401, opened 2018-03-20
Add FTVMCompute for initializers

nhynes:


kevinthesun:
Add some test cases to https://github.com/dmlc/nnvm/blob/master/tests/python/compiler/test_top_level1.py?

tqchen:
seems supported by #334

PR #405, opened 2018-03-21
CommonSubExpression Pass added

PariksheetPinjari909:
CommonSubExpression Pass added, Along with duplicate constants issue fixed in PrecomputePrune

### DIFF
--- a/src/compiler/common_sub_expression.cc
+++ b/src/compiler/common_sub_expression.cc
@@ -25,7 +25,7 @@ nnvm::Graph CommonSubExpression(nnvm::Graph src) {
       for (auto& e : node.node->inputs) {
         tempExprssn = find_cummulative_expression(e, tempExprssn);
         if (unique_expressn.count(tempExprssn)) {
-          // Replace already commoned node with current one
+          // Replace current one with already commoned node
           e = commonNodeList[unique_expressn.at(tempExprssn)];
           // Replace already commoned expression with its global index
           tempExprssn = std::to_string(unique_expressn.at(tempExprssn));
--- a/src/compiler/common_sub_expression.cc
+++ b/src/compiler/common_sub_expression.cc
@@ -14,14 +14,14 @@ nnvm::Graph CommonSubExpression(nnvm::Graph src) {
   std::vector<NodeEntry> commonNodeList;
   std::unordered_map<std::string, int> unique_expressn;
   std::string expressn = "";
-  std::function<std::string(const NodeEntry& e, std::string)>find_cummulative_expression
+  std::function<std::string(const NodeEntry& node, std::string)>find_cummulative_expression
     = [&find_cummulative_expression, &commonNodeList, &unique_expressn]
-    (const NodeEntry& e, std::string expressn)->std::string {
+    (const NodeEntry& node, std::string expressn)->std::string {
     std::string tempExprssn = "";
-    if (e.node->is_variable()) {
-      tempExprssn += e.node->attrs.name;
+    if (node.node->is_variable()) {
+      tempExprssn += node.node->attrs.name;
     } else {
-      for (auto& e : e.node->inputs) {
+      for (auto& e : node.node->inputs) {
         tempExprssn = find_cummulative_expression(e, tempExprssn);
         // Replace already commoned expression with its global index
         if (unique_expressn.count(tempExprssn)) {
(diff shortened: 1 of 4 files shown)


PariksheetPinjari909:
Common Sub Expression will basically take out common expression and define only once, 
for example:
 If the Graph has operation like:
   a = sqrt(y)
   b = sqrt(x)
   c = elementwise_add(a, b)
   d = sqrt(x)
   e = tanh(d)

Then after CommonSubExpression Pass applied the Graph will become as below:
   a = sqrt(y)
   b = sqrt(x)
   c = elementwise_add(a, b)
   e = tanh(b)    **## --> The common expression is gone**

tqchen:
@ZihengJiang can you do a codereview?

PariksheetPinjari909:
@ZihengJiang/ @Tqchen, request you to provide your valuable comments...or any issue regarding this feature... Thanks!

PariksheetPinjari909:
@ZihengJiang , @tqchen plz conclude on this pull request, thanks!

ZihengJiang:
Thanks for the PR. I will take a look tonight or tomorrow

src/compiler/common_sub_expression.cc
```diff
@@ -0,0 +1,57 @@
...
+        // Replace already commoned expression with its global index
+        if (unique_expressn.count(tempExprssn)) {
+          e = commonNodeList[unique_expressn.at(tempExprssn)];
+          tempExprssn = std::to_string(unique_expressn.at(tempExprssn));
+        }
+        expressn += tempExprssn;
+        tempExprssn = "";
+      }
+
+      tempExprssn = node.node->op()->name + expressn;
```
ZihengJiang: what if nodes have the same op but different attribute?

PariksheetPinjari909: @ZihengJiang , this is not handled in current code, i will modify code and update the PR. Thanks for the comment!

PariksheetPinjari909:
@ZihengJiang , your review comment is fixed now, would you please conclude on this PR, thanks!

PariksheetPinjari909:
@ZihengJiang , would you please conclude on this PR, thanks!

ZihengJiang:
Hi @PariksheetPinjari909 , in the PR, you find the common nodes by stringified operators, but seems I do not find any place to modify the graph. Correct me if I am wrong.

PariksheetPinjari909:
Hi @ZihengJiang , i have updated clear comments now where the current graph node is replaced with commoned node if macthes, @ Line no: 29, common_sub_expression.cc. 

Please let me know in case any other query/concern, Thanks!

PariksheetPinjari909:
@ZihengJiang , would you please conclude on this PR.

TaoLv:
Maybe it's a general question for all optimization passes. How many times will this pass be applied to an input graph? Can we expect that there will generate new CSEs after other optimization passes are applied?

src/compiler/common_sub_expression.cc
```diff
@@ -0,0 +1,63 @@
...
+ */
+#include <nnvm/graph.h>
+#include <nnvm/pass.h>
+
+namespace nnvm {
+namespace compiler {
+
+nnvm::Graph CommonSubExpression(nnvm::Graph src) {
+  std::vector<NodeEntry> commonNodeList;
+  std::unordered_map<std::string, int> unique_expressn;
```
TaoLv: I'm trying to understand this code. Seems you are mix-using `expression`, `expressn` and `exprssn`?

PariksheetPinjari909: Yes, please suggest one common term, will make it consistent.

In reply to your previous comment, CSE pass should apply only once in the input graph, it will modify the graph only if node exists which evaluates to same expressions, otherwise there wont be any change.

I hope this answer satisfis your query, please post if anything else. Thanks

TaoLv: Thanks, @PariksheetPinjari909 . 
Personally, I would like to use `expr` or full `expression`.
For CSE here, I guess it's similar to the `merge` optimization in Theano. `merge` optimization will be applied several times since the computation graph may be changed by other optimizations. 
http://www.deeplearning.net/software/theano/optimizations.html#term-merge 
http://www.deeplearning.net/software/theano/extending/optimization.html#optdb-structure

PariksheetPinjari909: @TaoLv , i have gone through the link you shared, there also it is mentioned that the Merge can be present only once that is at the end.

And for the same reason i have placed the CSE pass after Optimize() and before precomputeprune() pass, as there wont be any Graph modification which will cause the Nodes which are different before and post evaluate to same. I feel once the Graph is filtered with CSE Pass, there is no need to run again.
Please let me know if i miss anything.

tqchen:

PR #411, opened 2018-03-24
Add FCompute pass

nhynes:
This PR allows specifying nodes in terms of other nodes, which is helpful when trying to not rewrite FTVMComputes and FGradients.


include/nnvm/op_attr_types.h
```diff
@@ -151,6 +151,21 @@ using FInplaceIdentity = std::function<std::vector<bool> (const NodeAttrs& attrs
...
+/*!
+ * \brief Get the output node of the op node
+ *  This function generates the graph of the node
+ * \param nodeptr The current node
+ * \param inputs The current node's inputs
+ * \param input_shapes The current node's input shapes
+ * \return the Node's outputs
+ *
+ * \note Register under "FCompute"
+ */
```
tqchen: Rename to FExpandCompute, because fcompute was used by downstream for getting compute kernels

tqchen:
Please take a look at the CI error

python/nnvm/compiler/build_module.py
```diff
@@ -156,11 +156,13 @@ def optimize(graph, shape, dtype="float32"):
     cfg = BuildConfig.current
     if cfg.pass_enabled("SimplifyInference"):
         graph = graph_attr.set_shape_inputs(graph, shape)
-        graph = graph.apply(["InferShape", "SimplifyInference"])
+        graph = graph.apply(["InferShape", "ExpandCompute",
+                             "InferShape", "SimplifyInference"])
```
nhynes: it'd be nice to mark a pass as preserving an attribute so shape doesn't have to be inferred so often

tqchen: We are working on something toward that

src/compiler/expand_fcompute.cc
```diff
@@ -0,0 +1,48 @@
...
+#include <nnvm/top/nn.h>
+#include "./graph_transform.h"
+#include "./pattern_util.h"
+
+namespace nnvm {
+namespace compiler {
+
+Graph ExpandCompute(nnvm::Graph src) {
+  const IndexedGraph& idx = src.indexed_graph();
+  std::unordered_map<std::string, uint32_t> name2idx;
```
nhynes: this bit is required when invoking `ftransform` on a node's inputs requires a copy. being a copy, the input pointer doesn't exist in the indexed graph and can't be fetched using `idx.entry_id`.

src/compiler/expand_fcompute.cc
```diff
@@ -0,0 +1,64 @@
...
+  Graph egraph = GraphTransform(src, transform);
+  const IndexedGraph& eidx = egraph.indexed_graph();
+  for (int i = 0; i < eidx.num_nodes(); ++i) {
+    std::string name = eidx[i].source->attrs.name;
+    if (name2shape.count(name)) {
+      new_shape_vec.push_back(name2shape[name]);
+    } else {
+      new_shape_vec.emplace_back();
+    }
+  }
```
nhynes: this is incredibly brittle but recursive expansion requires that shape information be preserved

PR #416, opened 2018-03-26
Onnx opset support

fumihwh:
- Create `OnnxOpConvertor` class
- Create classes per op exists
- Support op level version control
- Fix incorrect representation (onnx.load get ModelProto not GraphProto, so return should be model)
- Update onnx version to 1.1.0


tqchen:
@zhreshold can you do a round of review?

python/nnvm/frontend/onnx.py
```diff
@@ -4,87 +4,113 @@
...
+        ]
+        versions = sorted(versions + [opset])
+        version = versions[
+            max([i for i, v in enumerate(versions) if v == opset]) - 1]
+        if hasattr(cls, '_impl_v{}'.format(version)):
+            return getattr(cls, '_impl_v{}'.format(version))
         else:
-            raise NotImplementedError("Only 2d kernel supported.")
-    return _impl
+            raise NotImplementedError(
```
zhreshold: Does it require rewriting _impl_vx for every onnx update?
How about a fallback support?

zhreshold:
@fumihwh So this PR is abandoning onnx models earlier than 1.1.0?

fumihwh:
@zhreshold 
No. This PR has compatibility with old models.

onnx version management is a little complex.
onnx version doesn't do anything with specifications of operators in practice. opset does.
But max version of opset in onnx 1.1.0 is 6? (maybe), we can say if we support onnx 1.1.0, we support opset to 6 from another point.

python/nnvm/frontend/onnx.py
fumihwh: @zhreshold 
If onnx updates special operator specification, we just need add one `_impl_` under that operator.
For example, if onnx add opset=7's Reshape op, we should add `_impl_v7` in class Reshape. Just that.

fumihwh:
@tqchen 
After merge master to branch, pylint gets angry with some missing-docstring.
I didn't change anything.....why...

tqchen:
we are upgrading the test docker env, which means the pylint might get upgraded

PR #421, opened 2018-03-30
[TOP] Adding compute for more operators

nhynes:
This is a rebase of #349


nhynes:
Actually, it seems like the only difference is that there's a bug in the current implement of `dense`. Tentatively closing.

PR #427, opened 2018-04-04
add support to `clip` operator

liangfu:


tqchen:
This will serve as temp solution, a better way might be introduce a clip operator. Note that topi support is already there

tqchen:
it appeared that clip is already added by #334

PR #437, opened 2018-04-11
[Quantization] Quantized Ops

ZihengJiang:


tqchen:
please fix the lint errors

liangfu:
I think quantization is an interesting topic, however, it requires large mount of effort to contribute and ensure everything works. **Can we make an independent branch on dmlc/nnvm, similar like [sparse](https://github.com/dmlc/nnvm/tree/sparse), so that more people can contribute to the topic?**

tqchen:
independent branch means divergence.  The quantization support likely does not need structure change so we will directly push for this effort in the master branch. With the help of TVM stack, the generation of quantized ops becomes easier and we already have some successful examples. 

We definitely welcome contribution and involvements. Either via issue, or discussion forum in tvm stack

python/nnvm/top/quantized_ops.py
```diff
@@ -0,0 +1,106 @@
...
+@reg.register_compute("quantize")
+def compute_quantize(attrs, inputs, _):
+    """Compute definition of quantize"""
+    repr_bit = attrs.get_int('repr_bit')
+    out_dtype = attrs['out_type']
+    assert out_dtype == 'int8'
+    data = inputs[0]
+
+    bit_width = 8
+    repr_value = pow(2, repr_bit)
```
tqchen: use left shift

python/nnvm/top/tensor.py
```diff
@@ -37,11 +36,42 @@ def _compute(attrs, x, _):
         return f(x[0], x[1])
     return _compute
 
-
 _fschedule_injective = tvm.convert(_schedule_injective)
 _fschedule_broadcast = _fschedule_injective
 _fschedule_elemwise = _fschedule_injective
 
+# lshift
+@reg.register_compute("left_shift")
```
tqchen: this is not consistent with numpy left_shift. To make it consistent with current add, sub, we should name it ```__lshift_scalar__```

python/nnvm/top/tensor.py
```diff
@@ -37,11 +36,42 @@ def _compute(attrs, x, _):
...
+def compute_lshift(attrs, inputs, _):
+    data = inputs[0]
+    bit = attrs.get_int("bit")
+    return topi.left_shift(data, bit)
+reg.register_pattern("left_shift", OpPattern.ELEMWISE)
+reg.register_schedule("left_shift", _fschedule_broadcast)
+
+# rshift
+@reg.register_compute("right_shift")
```
tqchen: this is not consistent with numpy left_shift. To make it consistent with current add, sub, we should name it ```__rshift_scalar__```

Laurawly:
Quantization is better to be built inside MXNet, and NNVM focuses on generating kernels. Usually in quantization process, fine-tune is required. Ideally, quantization is done in MXNet (with or without fine-tune), then use NNVM to generate lib.

tqchen:
This is mainly for graph level and op def support for quantized models, with simple support for graph re-transformation and do not intend to serve the role of training quantized models

liangfu:
The advantage of quantization within NNVM over MXNet, I think, is that it provides accelerated computation other than CUDA, typically for devices that support OpenCL like AMD GPUs and FPGAs. Besides, well designed neural network should be robust to quantized computation in nature. For instance, the introduction of `relu6` over `relu` would reduce the chance of numeric overflow, while still resist to degenerate accuracy.

include/nnvm/top/tensor.h
```diff
@@ -251,6 +265,15 @@ struct ClipParam : public dmlc::Parameter<ClipParam> {
   }
 };
 
+struct AroundParam : public dmlc::Parameter<AroundParam> {
+  int mode;
+  DMLC_DECLARE_PARAMETER(AroundParam) {
+    DMLC_DECLARE_FIELD(mode);
```
etaf: This is a typo? It cause compilation failure. Suggest to mv the ";" to the end of next line.

etaf:
@ZihengJiang May I know whether this PR will be merged?

tqchen:

Issue #303, opened 2018-01-12
[RFC][PASS] Weight Prepack


tqchen:
In many scenarios, we will need to run computation under a different data layout, in such cases, it is desirable to precompute the weight in a packed format. Note that this is a restricted case of layout optimization, which we see a few times now. So I propose to have a pass to specifically only solve this case, to limit the scope of the problem. This won't conflict with other layout optimization passes.

### Necessary steps
- [ ] Specialized conv2d operator on contrib that corresponds to a typical algorithm, something like spatial_conv2d, im2col_conv2d, with flags indicating whether weight is packed.
   - These are intermediate artifacts of the graph optimizer that is not part of core op spec.
- [ ] Add register_weight_prepack to https://github.com/dmlc/nnvm/blob/master/src/compiler/packed_func_ext.cc#L84
     - With signature (NodeAttrs, Symbol children, Array<Tensor> tinfo) (note that we can pass children as symbol)  
          - tinfo are placeholders that to store shape and type information of each child.
     - When registered, can return Symbol of transformed weight,
```python
def transform_conv2d(attrs, children, shape_info):
     w = children[1]
     # can use nnvm.transpose and nnvm.reshape to pack things up.
     w = pack(w)
     return nnvm.contrib.spatial_pack_conv2d(children[0], w, pack_w=True)
```
- [ ] Register compute schedules for the specific ops.

tqchen:
@ZihengJiang @merrymercy

yzhliu:
You can assign it to me, will shoot a PR soon.

merrymercy:
@yzhliu any update?

yzhliu:
@merrymercy cleaning the code. will PR on Tuesday or Wednesday.

merrymercy:
@yzhiliu  Besides bijective layout transform, we still need some specialized weight transform. For example, winograd and FFT.

@tqchen  If following your steps, where will we call `transform_conv2d`? Shall we also add a new pass?

yzhliu:
If you do not want to leverage the automatic layout transform, especially for weight packing only, you can simply ignore `InferCorrectLayout` pass.

merrymercy:
I see. It is good.

yzhliu:
Looks like this issue can be closed. @tqchen

PR #435, opened 2018-04-09
update list of operators defined in nnvm

liangfu:

PR #447, opened 2018-04-18
General Layout Support

yzhliu:
This PR allows NNVM to do
* Replace an operator via `AlterOpLayout`. For x86 it is more efficient to calculate convolution in `NCHW16c` layout. Here is an example of how it is used: https://github.com/yzhliu/topi-intel/blob/master/e2e_general_pack/schedule_pack/avx512_conv_fwd.py#L128-L155
Note that kernel (pre-)packing is supported as well: https://github.com/dmlc/nnvm/issues/303

* Infer and correct layout automatically
  - Given the input data layout, or the layout of operators in the network (e.g., convolution, pooling...) `InferCorrectLayout` pass can infer the layout for each operator (both inputs and outputs).
  - If the required input layout of an operator is different from what it receives, a `__layout_transform__` operator will be inserted.
  - Each operator registers a function `FInferLayout`. Once a model is imported,  we do a `InferCorrectLayout` pass and store the *original layouts* for each operator. After `AlterOpLayout`, the `InferCorrectLayout` runs again. This time each operator will see the *original layouts* it inferred before, which can be used to decide whether to keep the original one. For example, `softmax` can still generate correct result after the input layout changes, while `flatten` cannot. So `flatten` claims it needs the `original layout` and triggers a layout transform, which make the network produce correct result.
  - With such approach,
     * optimized layout (e.g., NCHW16c) can flow through the network as far as possible. No pack/unpack overhead for each (e.g., convolution) operator.
     * now model layout can be transparent to users. Even though a convolution neural network is trained with `NHWC` layout, users can still pass `NCHW` input as long as the input layout is specified, a layout transform happens automatically.

Moreover the convolution kernel layout becomes much clearer: https://github.com/dmlc/nnvm/pull/372 . Now we have kernel layout `OIHW` and `HWIO` for `NCHW` and `NHWC` respectively.

Will shoot a pull request for corresponding TVM changes.

@yidawang @kevinthesun


python/nnvm/compiler/build_module.py
```diff
@@ -204,8 +218,8 @@ def build(graph, target=None, shape=None, dtype="float32", params=None, target_h
         By default, llvm is used if it is enabled,
         otherwise a stackvm intepreter is used.
 
-    initialize : bool, optional
-        Whether to initialize variables in global dict _all_var_init.
+    layout : dict of str to str or str
```
kevinthesun: Does user need to specify opt_level=3 to enable layout transform? Also how does user specify "internal" layout, if the target layout is not NCHWXc?

src/top/elemwise_op_common.h
```diff
@@ -100,12 +103,176 @@ inline bool ElementWiseReduceType(const NodeAttrs& attrs,
...
+          << name << ": " << "expected " << *target
+          << ", got " << vec->at(i);
+      }
+    }
+  };
+
+  Layout in, last_in, out;
+  deduce(&in, in_layouts, in_size, "input");
+  deduce(&last_in, last_in_layouts, in_size, "input (last infer pass)");
+  deduce(&out, out_layouts, out_size, "output");
```
kevinthesun: Assume an elemwise_add op has two inputs with layout "NCHW16c" and "NCHW4c", will they be transformed back to "NCHW", or one of them will be transformed to another?

yzhliu: right transform to left. it's defined in `ElemwiseBinaryKeepLeftLayout`

python/nnvm/frontend/mxnet.py
```diff
@@ -237,7 +247,6 @@ def _upsampling(inputs, attrs):
     'min_axis'      : _rename('min'),
     'reshape'       : _reshape,
     'sum_axis'      : _rename('sum'),
-    'UpSampling'    : _upsampling
```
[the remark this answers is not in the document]
yzhliu: oh I was removing SSD stuff but by mistake remove this...

include/nnvm/compiler/packed_func_ext.h
```diff
@@ -52,6 +54,12 @@ template<>
 struct extension_class_info<nnvm::compiler::AttrDict> {
   static const int code = 18;
 };
+
+template<>
+struct extension_class_info<nnvm::compiler::SymbolArray> {
+  static const int code = 19;
```
tqchen: cannot use 19, because it is taken by mxnet tvm bridge https://github.com/apache/incubator-mxnet/blob/master/include/mxnet/tensor_blob.h#L49

include/nnvm/compiler/op_attr_types.h
```diff
@@ -73,19 +74,17 @@ using FTVMSchedule = std::function<
...
+ * \param attrs The attribute of the original node.
+ * \param inputs The input symbols of the original node.
+ * \param tinfos The inferred shape and dtype of the inputs.
  */
-using FTVMLayoutRequest = std::function<bool (const NodeAttrs& attrs,
-                                              std::vector<TLayoutInfo> *ilayouts,
-                                              std::vector<TLayoutInfo> *olayouts)>;
+using FTVMAlterOpLayout = std::function<
+  Symbol(const NodeAttrs& attrs,
+         const SymbolArray& inputs,
```
tqchen: do we really need the input symbols, or is tinfo enough for now?

include/nnvm/compiler/packed_func_ext.h
tqchen: It seems to be quite dangerous to expose a vector of pointers here.

include/nnvm/compiler/op_attr_types.h
```diff
@@ -73,19 +74,17 @@ using FTVMSchedule = std::function<
...
+ *  it is invoked in AlterOpLayout pass.
+ * \param attrs The attribute of the original node.
+ * \param inputs The input symbols of the original node.
+ * \param tinfos The inferred shape and dtype of the inputs.
  */
-using FTVMLayoutRequest = std::function<bool (const NodeAttrs& attrs,
-                                              std::vector<TLayoutInfo> *ilayouts,
-                                              std::vector<TLayoutInfo> *olayouts)>;
+using FTVMAlterOpLayout = std::function<
+  Symbol(const NodeAttrs& attrs,
```
tqchen: Can we just return the result output layout as vector of string?

include/nnvm/compiler/op_attr_types.h
tqchen: ignore this comment, I now get what is going on

src/pass/infer_correct_layout.cc
```diff
@@ -0,0 +1,171 @@
...
+  return n;
+}
+
+using LayoutAttrDict = std::unordered_map<const Node*, std::vector<Layout> >;
+
+/*!
+ * \brief A simple layout infer pass that will
+ *        insert layout transform nodes automatically.
+ */
+nnvm::Graph InferCorrectLayout(nnvm::Graph src) {
```
tqchen: Maybe we can just call it CorrectLayout

python/nnvm/top/symbol_array.py
```diff
@@ -0,0 +1,35 @@
+# pylint: disable=invalid-name
+"""Symbol array object used by weight pre-pack"""
+import tvm
+
+_sym_arr_get = tvm.get_global_func("nnvm.compiler._symbol_array_get")
+_sym_arr_size = tvm.get_global_func("nnvm.compiler._symbol_array_size")
+
+class SymbolArray(object):
```
tqchen: We can just use symbol for this, because we can group multiple nodes into a single Symbol

include/nnvm/compiler/op_attr_types.h
tqchen: ignore this

src/pass/infer_correct_layout.cc
yzhliu: ok. will change.

include/nnvm/layout.h
```diff
@@ -0,0 +1,455 @@
...
+#include <sstream>
+#include <vector>
+#include <utility>
+#include <algorithm>
+
+namespace nnvm {
+
+class Layout {
+ public:
+  using LayoutDim = char;
```
yzhliu: @tqchen 
Do you want me to change the name 'dim' to 'axis' as you suggested in https://discuss.tvmlang.org/t/datalayout-structure/80/3 ? 

And to my understanding, we don't have to wait for that DataLayout implemented in TVM. We can merge this PR first (so that we can move forward to having significant improvement - at least 50% speedup - for CNN) and switch to that in TVM later, right?

tqchen: yes, we can merge this in first, please fix the problems I mentioned

tqchen:
@merrymercy Can you also do a review when you have time?

yzhliu:
@tqchen The segment fault in CI looks weird. I cannot reproduce. And it happened in PR #435 as well: http://mode-gpu.cs.washington.edu:8080/blue/organizations/jenkins/dmlc%2Fnnvm/detail/PR-435/13/pipeline

src/pass/correct_layout.cc
```diff
@@ -0,0 +1,171 @@
...
+    for (uint32_t i = 0; i < inode.inputs.size(); ++i) {
+      const auto& e = inode.inputs[i];
+      const nnvm::NodePtr& in = mirror_vec[e.node_id];
+      new_node->inputs[i] = nnvm::NodeEntry{in, e.index, e.version};
+
+      // insert layout_transform if necessary
+      const Layout& produce = produce_ilayouts[i];
+      const Layout& request = request_ilayouts[i];
+      if (produce != request && produce.defined()) {
+        LOG(INFO) << "Insert layout transformer for " << i << "-th input of "
```
tqchen: This seems to produce a lot of messages, maybe avoid print this, since this is expected behavior of the pass

tqchen:
Thanks for improving the code through the review! this is now merged

PR #468, opened 2018-05-06
input shape and padding scheme based padding caluation support (part 2/2)

srkreddy1238:
* Compatible to 'SAME' option across different frameworks.


tqchen:
First of all the topi's conv2d already support SAME/VALID as input. The current NNVM convention favor explicit padding values. We usually need to take great caution when updating the spec of the operator. In this case it may not be necessary

We can explicitly  calculate the padding shape based on input shape during conversion. Please refer to keras frontend on how to do so.

srkreddy1238:
Thanks for the info, I will rework on the frontend to use the pad operator instead.

PR #472, opened 2018-05-09
[REVIEW] Tensorflow frontend support

srkreddy1238:
Inception V1 supported now.

### DIFF
--- a/python/nnvm/frontend/__init__.py
+++ b/python/nnvm/frontend/__init__.py
@@ -5,3 +5,4 @@ from .onnx import from_onnx
 from .coreml import from_coreml
 from .keras import from_keras
 from .darknet import from_darknet
+from .tensorflow import from_tensorflow
--- a/python/nnvm/frontend/tensorflow.py
+++ b/python/nnvm/frontend/tensorflow.py
@@ -1,10 +1,6 @@
-# pylint: disable=import-self, invalid-name, unused-argument, too-many-nested-blocks, no-else-return, line-too-long
+# pylint: disable=import-self, invalid-name, unused-argument
 """TF: Tensorflow frontend."""
 from __future__ import absolute_import as _abs
-
-# Tensorflow imports
-from tensorflow.python.framework import tensor_util
-from tensorflow.python.framework import dtypes
 
 # Numpy support
 import numpy as np
@@ -33,19 +29,23 @@ class AttrCvt(object):
 
     def __call__(self, inputs, attrs, *args):
         self._ignores.append('_output_shapes')
+        self._ignores.append('_input_shapes')
         self._ignores.append('T')
         self._ignores.append('use_cudnn_on_gpu')
         return AttrConvert(self._op_name, self._transforms, self._excludes,
-                           self._disables, self._ignores, self._extras, self._custom_check)(inputs, attrs, *args)
+                           self._disables, self._ignores, self._extras,
+                           self._custom_check)(inputs, attrs, *args)
 
-def _get_input_shapes(attr):
-    return [tensor_util.TensorShapeProtoToList(shape) for shape in attr['_output_shapes']]
+def _get_pad_pair(input1d, kernel1d, stride1d):
+    if input1d % stride1d == 0:
+        pad = max(kernel1d - stride1d, 0)
+    else:
+        pad = max(kernel1d - (input1d % stride1d), 0)
 
-def _get_pad(input1d, kernel1d, stride1d):
-    out1d = (input1d + stride1d - 1) // stride1d
-    pad = np.maximum((out1d - 1) * stride1d + kernel1d - input1d, 0)
-    pad = pad // 2
-    return pad
+    pad_before = pad // 2
+    pad_after = pad - pad_before
+
+    return [pad_before, pad_after]
 
 def _math_name_picker(surfix):
     def _impl(attr):
@@ -69,7 +69,7 @@ def _dimension_constraint():
     return _dim_check, "Only 2d kernel supported."
 
 def _infer_channels(inputs, params, transpose=False):
-    """A hack for getting 'channles' or 'units' since onnx don't provide
+    """A hack for getting 'channles' or 'units' since tensorflow don't provide
     these attributes. We check the shape of weights provided to get the number.
     """
     g = _graph.create(inputs)
@@ -97,14 +97,16 @@ def _pooling(name):
 
         if attr['data_format'] == 'NHWC':
             attr['kernel_shape'] = (attr['ksize'][1], attr['ksize'][2])
-        if attr['data_format'] == 'NCHW':
+        elif attr['data_format'] == 'NCHW':
             attr['kernel_shape'] = (attr['ksize'][2], attr['ksize'][3])
+        else:
+            raise TypeError("Unsupported data_format type : {}".format(attr['data_format']))
 
         # Fix strides
         attr['strides'] = (attr['strides'][1], attr['strides'][2])
 
         # Fix padding
-        input_shapes = _get_input_shapes(attr)
+        input_shapes = attr['_input_shapes'][inputs[0]]
         attr['padding'] = attr['padding'].decode("utf-8")
 
         if attr['padding'] == 'VALID':
@@ -112,11 +114,29 @@ def _pooling(name):
         elif attr['padding'] == 'SAME':
             stride_h, stride_w = attr['strides']
             kernel_h, kernel_w = attr['kernel_shape']
-            in_h = input_shapes[0][1]
-            in_w = input_shapes[0][2]
-            pad_t = _get_pad(in_h, kernel_h, stride_h)
-            pad_l = _get_pad(in_w, kernel_w, stride_w)
-            attr['padding'] = [pad_t, pad_l]
+            if attr['data_format'] == 'NHWC':
+                in_h = input_shapes[0][1]
+                in_w = input_shapes[0][2]
+            else:
+                in_h = input_shapes[0][2]
+                in_w = input_shapes[0][3]
+
+            pad_v = _get_pad_pair(in_h, kernel_h, stride_h)
+            pad_h = _get_pad_pair(in_w, kernel_w, stride_w)
+
+            if attr['data_format'] == 'NHWC':
+                inputs[0] = _sym.pad(data=inputs[0],
+                                     pad_width=((0, 0),
+                                                (pad_v[0], pad_v[1]),
+                                                (pad_h[0], pad_h[1]),
+                                                (0, 0)))
+            else:
+                inputs[0] = _sym.pad(data=inputs[0],
+                                     pad_width=((0, 0),
+                                                (0, 0),
+                                                (pad_v[0], pad_v[1]),
+                                                (pad_h[0], pad_h[1])))
+            attr['padding'] = [0, 0]
         else:
             raise TypeError("Unsupported padding type : {}".format(attr['padding']))
 
@@ -142,17 +162,19 @@ def _conv():
             attr['channels'] = conv_param_weights.shape[3]
             if 'dilations' in attr:
                 attr['dilations'] = (attr['dilations'][0], attr['dilations'][1])
-        if attr['data_format'] == 'NCHW':
+        elif attr['data_format'] == 'NCHW':
             attr['kernel_shape'] = (conv_param_weights.shape[2], conv_param_weights.shape[3])
             attr['channels'] = conv_param_weights.shape[1]
             if 'dilations' in attr:
                 attr['dilations'] = (attr['dilations'][2], attr['dilations'][3])
+        else:
+            raise TypeError("Unsupported data format type : {}".format(attr['data_format']))
 
         # Fix strides
         attr['strides'] = (attr['strides'][1], attr['strides'][2])
 
         # Fix padding
-        input_shapes = _get_input_shapes(attr)
+        input_shapes = attr['_input_shapes'][inputs[0]]
         attr['padding'] = attr['padding'].decode("utf-8")
 
         if attr['padding'] == 'VALID':
@@ -160,11 +182,30 @@ def _conv():
         elif attr['padding'] == 'SAME':
             stride_h, stride_w = attr['strides']
             kernel_h, kernel_w = attr['kernel_shape']
-            in_h = input_shapes[0][1]
-            in_w = input_shapes[0][2]
-            pad_t = _get_pad(in_h, kernel_h, stride_h)
-            pad_l = _get_pad(in_w, kernel_w, stride_w)
-            attr['padding'] = [pad_t, pad_l]
+            if attr['data_format'] == 'NHWC':
+                in_h = input_shapes[0][1]
+                in_w = input_shapes[0][2]
+            else:
+                in_h = input_shapes[0][2]
+                in_w = input_shapes[0][3]
+
+            pad_v = _get_pad_pair(in_h, kernel_h, stride_h)
+            pad_h = _get_pad_pair(in_w, kernel_w, stride_w)
+
+            if attr['data_format'] == 'NHWC':
+                inputs[0] = _sym.pad(data=inputs[0],
+                                     pad_width=((0, 0),
+                                                (pad_v[0], pad_v[1]),
+                                                (pad_h[0], pad_h[1]),
+                                                (0, 0)))
+            else:
+                inputs[0] = _sym.pad(data=inputs[0],
+                                     pad_width=((0, 0),
+                                                (0, 0),
+                                                (pad_v[0], pad_v[1]),
+                                                (pad_h[0], pad_h[1])))
+
+            attr['padding'] = [0, 0]
 
         else:
             raise TypeError("Unsupported padding type : {}".format(attr['padding']))
@@ -201,12 +242,13 @@ def _expand_dims():
         dim_input = inputs.pop(1)
         axis = params[dim_input.list_output_names()[0]]
         params.pop(dim_input.list_output_names()[0])
-        return AttrCvt(op_name="expand_dims", ignores=['Tdim'], extras={'axis': axis.asnumpy()[0]})(inputs, attr)
+        return AttrCvt(op_name="expand_dims", ignores=['Tdim'],
+                       extras={'axis': axis.asnumpy()[0]})(inputs, attr)
     return _impl
 
 def _resize_bilinear():
     def _impl(inputs, attr, params):
-        # TODO: Making a copy node assuming the input image shape is 299x299
+        # Making a copy node assuming the input image shape is 299x299
         # Change this when we have corresponding resize bilinear operation.
         pop_node = inputs.pop(1)
         params.pop(pop_node.list_output_names()[0])
@@ -215,7 +257,7 @@ def _resize_bilinear():
 
 def _check_numerics():
     def _impl(inputs, attr, params):
-        # TODO: Making a copy node assuming no need to verify
+        # Making a copy node assuming no need to verify
         return AttrCvt(op_name="copy", ignores=['message'])(inputs, attr)
     return _impl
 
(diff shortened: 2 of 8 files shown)


srkreddy1238:
@tqchen review pls

tqchen:
@Huyuwei can you help do a pass of review?

python/nnvm/frontend/__init__.py
```diff
@@ -5,3 +5,4 @@
 from .coreml import from_coreml
 from .keras import from_keras
 from .darknet import from_darknet
+from .tensorflow import from_tensorflow
```
tqchen: frontend should always be optional, this means unless from_tensorflow is called, we should not depend on availability of tensorflow

python/nnvm/frontend/tensorflow.py
```diff
@@ -0,0 +1,512 @@
+# pylint: disable=import-self, invalid-name, unused-argument, too-many-nested-blocks, no-else-return, line-too-long
```
tqchen: do not disable long-to-long, no-else-return too-many-nested-blocks,

python/nnvm/frontend/__init__.py
tqchen: pls refer to the implementation of from_mxnet etc to see how do they import the dependency within the function

tqchen:
@FrozenGene would you be interested in review the code?

Huyuwei:
Sure, will do this weekend.

FrozenGene:
@tqchen Thanks for invitation. I am doing the padding support as we discussed in the forum(https://discuss.tvmlang.org/t/tvm-seems-conv2d-has-relate-big-floating-point-error-when-to-from-coreml/145/13). My implementation of Conv2D and depthwise2D has completely exclude scheduling part. And we can have the same result as Tensorflow, CoreML. My test model is MobileNet, SSD MobileNet, the model files is from tensorflow official repo and convert into CoreML format, then use NNVM to support it. Currently, I just review the padding part, but I find this implementation is also not contain (top, left, bottom, right) 4 elements, so I think this implementation should also have the same problem of padding.

python/nnvm/frontend/tensorflow.py
```diff
@@ -0,0 +1,512 @@
...
+
+def _dimension_constraint():
+    def _dim_check(attrs):
+        if len(attrs['kernel_shape']) == 2:
+            return True
+        return False
+    return _dim_check, "Only 2d kernel supported."
+
+def _infer_channels(inputs, params, transpose=False):
+    """A hack for getting 'channles' or 'units' since onnx don't provide
```
Huyuwei: not onnx

python/nnvm/frontend/tensorflow.py
```diff
@@ -0,0 +1,512 @@
...
+        # Fix strides
+        attr['strides'] = (attr['strides'][1], attr['strides'][2])
+
+        # Fix padding
+        input_shapes = _get_input_shapes(attr)
+        attr['padding'] = attr['padding'].decode("utf-8")
+
+        if attr['padding'] == 'VALID':
+            attr['padding'] = [0, 0]
+        elif attr['padding'] == 'SAME':
```
Huyuwei: I am afraid SAME padding is not correctly converted here.
Can you add a test with input_h=32, kernel_h=3, stride_h=2?

python/nnvm/frontend/tensorflow.py
```diff
@@ -0,0 +1,512 @@
...
+    def _impl(inputs, attr, params):
+        return _sym.broadcast_add(inputs[0], inputs[1])
+    return _impl
+
+def _batch_norm():
+    def _impl(inputs, attr, params):
+        # Rearrange inputs from
+        # (data, moving_mean, moving_variance, beta, gamma) to
+        #     to
+        # (data, gamma, beta, moving_mean, moving_var)
```
Huyuwei: to to

python/nnvm/frontend/tensorflow.py
```diff
@@ -0,0 +1,512 @@
...
+# for 1 to 1 mapping, use Renamer if nothing but name is different
+# use AttrCvt if attributes need to be converted
+# for 1 to N mapping(composed), use custom callable functions
+# for N to 1 mapping, currently not supported(?)
+_convert_map = {
+    'AvgPool'                           : _pooling('avg_pool'),
+    'BatchNormWithGlobalNormalization'  : _batch_norm(),
+    'BiasAdd'                           : _bias_add(),
+    'Cast'                              : _cast(),
+    'CheckNumerics'                     : _check_numerics(),                      # TODO
```
Huyuwei: TODO here is confusing. Maybe comment out unsupported operators?

python/nnvm/frontend/tensorflow.py
```diff
@@ -0,0 +1,512 @@
...
+    'Relu'                              : AttrCvt('relu'),
+    'Reshape'                           : _reshape(),
+    'ResizeBilinear'                    : _resize_bilinear(),                     # TODO
+    'Softmax'                           : AttrCvt('softmax', {'axis': ('axis', 1)}),
+    'Sub'                               : _elemwise('sub'),
+}
+
+
+class GraphProto(object):
+    """ TODO: A helper class for handling nnvm graph copying from Tensorflow GraphDef.
```
Huyuwei: TODO here is confusing, since GraphProto is implemented.

python/nnvm/frontend/tensorflow.py
```diff
@@ -0,0 +1,512 @@
...
+        self._renames = {}
+        self._replacements = {}
+        self._output_shapes = {}
+        self._num_input = 0
+        self._num_param = 0
+        self._input_node = ''
+
+    def from_tensorflow(self, graph):
+        """Construct nnvm nodes from tensor flow  graph definition - GraphDef.
+        TODO: Detailed explanation of TF GraphDef parsing.
```
Huyuwei: Add the explanation and remove TODO.

tutorials/from_tensorflow.py
```diff
@@ -0,0 +1,206 @@
...
+predictions = np.squeeze(predictions)
+
+# Creates node ID --> English string lookup.
+node_lookup = NodeLookup()
+
+top_k = predictions.argsort()[-10:][::-1]
+for node_id in top_k:
+    human_string = node_lookup.id_to_string(node_id)
+    score = predictions[node_id]
+    print('%s (score = %.5f)' % (human_string, score))
```
Huyuwei: We can compare TVM output with tensorflow output to show the correctness.

Huyuwei:
@srkreddy1238 Tests should be added.

And many TODO are quite confusing. I think it's helpful to add some comments in the file to describe what is not yet supported in the tensorflow converter.

python/nnvm/frontend/tensorflow.py
FrozenGene: @Huyuwei Aha, this keras's implementation is interesting. But how to we handle the workload / schedule part? Because in our workload / schedule part does only concern VALID padding and only care top / left. see: https://github.com/dmlc/tvm/blob/9175b47c16691824b1a54aa26dc1ddb21c981612/topi/python/topi/nn/conv2d.py#L146 Though Keras's implementation solve SAME padding, but not solve pass this padding information to the scheduling part. This is why I discuss with @tqchen in https://discuss.tvmlang.org/t/tvm-seems-conv2d-has-relate-big-floating-point-error-when-to-from-coreml/145. @tqchen suggest extending Conv2D's operator's param(maybe include pooling) form [top, left] to [top, left, bottom, right]. So I want to know this Keras's implementation. how to pass the 4 padding information to scheduling? If we can solve this problem, we don't need to extend operator's param.

srkreddy1238: Initially I borrowed the pad_pair calculation from keras which I replaced with tensorflow reference.
Both work fine with InceptionV1 as long as I don't use pad operator before convolution, but pass pad_t, pad_l to conv2d.

I see the below error with pad operator. I am investigating it.

nnvm._base.NNVMError: Error in operator conv2d2: [14:09:41] src/top/nn/convolution.cc:84: Operator conv2d(use_bias=False, strides=(1, 1), channels=64, layout=NHWC, padding=[0, 0], kernel_size=(3, 3), kernel_layout=HWIO, name=conv2d2) expects weight's shape to be [3,3,34,64], but got [3,3,32,64].

python/nnvm/frontend/tensorflow.py
srkreddy1238: Done

python/nnvm/frontend/tensorflow.py
srkreddy1238: Simplified the TODO with one section explaining the implementation, limitation and assumption aspects.

python/nnvm/frontend/tensorflow.py
srkreddy1238: Simplified the TODO with one section explaining the implementation, limitation and assumption aspects.

python/nnvm/frontend/tensorflow.py
srkreddy1238: Simplified the TODO with one section explaining the implementation, limitation and assumption aspects.

tutorials/from_tensorflow.py
srkreddy1238: Sure I will add tensorflow prediction logic and do a comparision.

Also working on the test cases for TF.

python/nnvm/frontend/tensorflow.py
srkreddy1238: Above error due to different order of padding values for NHWC. 

Now the padding scheme for TF front end follows tensorflow reference with pad operator before convolution and it work fine with InceptionV1.

Huyuwei: @FrozenGene Now keras frontend inserts a pad operator before conv to handle SAME padding, and I expect the pad operator to be fused. Consider this sequence: extra_pad_1, pad_1, conv_1, relu_1, extra_pad_2, pad_2, conv2... relu_1 and extra_pad_2 will all be fused into conv_1. But I just noticed that extra_pad_2 won't be fused in the generated code. So this pad operator brings overhead.

I agree that we can extend conv2d's padding param to support [top, left, bottom, right]. It will make the conversion much easier and clearer. The pad API in keras may be helpful: https://keras.io/layers/convolutional/#zeropadding2d

FrozenGene: @Huyuwei Yes, but it will need much work in the scheduling part, because we don't consider bottom / right padding value in our previous design, I am working on it. And we will also need to extend conv2d_transpose / max_pool2d / avg_pool2d / global_max_pool2d / global_avg_pool2d. i.e. we need to extend all operators of Level 2 Convolutions in this table: http://nnvm.tvmlang.org/top.html#overview-of-operators

srkreddy1238: Another issue I came across with padding on pool operation while doing test cases for tensorflow frontend https://github.com/dmlc/nnvm/issues/486

I think we could move this topic to discussion forum.

srkreddy1238:
Done with all review comments listed below.
    Test cases added.
    Pad operator similar to keras.
    Corrections on internal documentation and coding standards.

Left:
Padding discussion, which is common for multiple frameworks.

python/nnvm/frontend/tensorflow.py
FrozenGene: @srkreddy1238 @Huyuwei I have extended padding completely and can have the same result as Tensorflow / CoreML now. However, I keep the implementation of padding [top, left], which is useful for the compatibility of old code. For example, we have unittest test_infer_shape using old behaviour. But if user specify [top, left, bottom, right], we go to the new behaviour. Another reason I keep the implementation is I find many framework(keras, MXNet and so on) are using padding[top, left], I don't have time to test every framework(Especially MXNet, I find its official documentation says that his padding parameter is only[top, left] too). But I think we should only have padding[top, left, bottom, right]. Do you have any ideas or suggestions?

srkreddy1238: Sounds good. Can you PR it to review ?

srkreddy1238:
@FrozenGene & @Huyuwei  any comments ??

srkreddy1238:
@FrozenGene 

(Whether we should keep old behaviour's compatibility?) : I feel we should keep the compatibility and ensure the support for all frameworks. 

The front end  for tensorflow is the initial work with minimal support which need to be evolved over time inline with TVM/NNVM evolving.

Here answering  your reviews.

Output nodes: Unlike other frameworks tensorflow protobuf I don't see a definition for inputs and output nodes specifically, hence assumed first node as input and last node as output. I have this multiple output support in mind, I am working on it.

BilinearResize is dummy as there is no supported topi. This can be fixed once we have the topi support in near future. I will correct the comment "299x299".

Tensorflow operator list is too huge compared to other frameworks, so I have choose to list supported list. I have plans to create categorized lists in coming PR for tensorflow frontend.

This is initial effort towards supporting tensorflow, support I am already listing on the missing pieces which include above issues to bring up with a plan.

python/nnvm/frontend/tensorflow.py
srkreddy1238: 👍 . I will help to check tensorflow once you PR it.

FrozenGene:
yes, we don't have BilinearResize op. I list one ref link that CoreML doesn't have BilinearResize op too, but it uses upsampling to do the work. So, I think if we don't BilinearResize op to done the mapping, we should do the work like CoreML's way, not just copy, which is not correct obviously.

srkreddy1238:
I see CoreML approach, but NNVM don't have operator to support any resize now.

DecodeJpeg is removed completely as we decided not to support decode in NNVM in anyway. I kept ResizeBilinear with a copy now assuming we come up with a topi for resize operator soon.

I will add a notify print in frontend to notify user about ResizeBilinear dummy functionality until we don't have correct implementation in place.

Feel free to comment on this.

FrozenGene:
We have UpSampling operator but don't have Bilinear mode. If you want to do like CoreML's approach, you need to add Bilinear mode's support : https://github.com/dmlc/tvm/blob/785c9420c1604dbaf85195c31be625f496e5776f/topi/python/topi/nn/upsampling.py Keras's implementation have also used this op: https://github.com/dmlc/nnvm/blob/master/python/nnvm/frontend/keras.py#L289

Or you should use some information to notify user as you said.

srkreddy1238:
Great, I will update to use this for time being until we have bilinear.

srkreddy1238:
Up_sample takes only scale as parameter. It's a symmetric resize of height & width. I doubt if this really serve the purpose even if we use here.

As we disabled the JpegDecode anyway user has to fix an input decoded image size for NNVM compilation, which the user could fix (ex 299 for inception v1) to the output of ResizeBilinear by handling resize in preprocessing. 

I feel ResizeBilinear with out DecodeJpeg may not make much sense here.

What do you think @FrozenGene ?

FrozenGene:
Yeah, I know UpSampling only accept scale now. As CoreML's approach, we can compute the scale factor, right? However, we lack of BILINEAR mode for UpSampling, as previous comment said, if you use this op, **you should exteand UpSampling op and add support of BILINEAR mode**, currently , we only have NN(Nearest Neighbors)mode for UpSampling.

If you think we should have resize_bilinear op to mapping and not using this op, you could also add notify information that we don't implement resize_bilinear right now and waiting NNVM have resize_bilinear op.

srkreddy1238:
Sure, I will add those changes today

srkreddy1238:
@FrozenGene  Done with upsampling in ResizeBilinear.

FrozenGene:
What I means in fact you should add Bilinear mode support in UpSampling (For example, accept one param named mode, then pass ‘Bilinear’ to UpSampling, we compute it intenally implementation of UpSampling. To keep old code working, we make ‘NN’ be default mode), not use UpSampling directly. Because ResizeBilinear maps to UpSampling op with Bilinear mode, not NN mode.

srkreddy1238:
Actually I am already working on ResizeBilinear core logic to use in upsampling or a separate operator which can be discussed. I will consider the new arg to differentiate Bilinear with in upsampling.

I wanted to let frontend go for merge to allow others try with different model and feedback. 

What do you suggest ?

FrozenGene:
I think ResizeBilinear maps to UpSampling with Bilinear directly, not NN mode with UpSampling. If you decide to check in code with ResizeBilinear support, we should make sure its support is right, currently   NN mode of UpSampling is not right obviously. We should support op right or leave it blank.

srkreddy1238:
I agree with you on support the right way if we quote Bilinear is supported.

I planned not to support Bilinear in initial version of frontend, so I made it dummy.
But I have plan to add it soon (may be in a week time).

Reason I was pushing the frontend is to let others try this and feedback. Also there is other work dependent on this front end.

@FrozenGene  and @tqchen pls advice.

FrozenGene:
Ok, I understand your point now. I think you could add ``` # Change this when we have corresponding resize bilinear operation.``` to our print, like this:
```python
print ("ResizeBilinear: Only nearest neighbor supported in symetric mode of dimensions")
print( "Change this when we have corresponding resize bilinear operation")
```

srkreddy1238:
Done with above change in the message.

FrozenGene:
@srkreddy1238 FYI. https://github.com/dmlc/tvm/pull/772. The author said Bilinear mode could be done with conv2d_transpose with certain weight. When you implement it, you could talk with him. Or you can persuade him UpSampling should accept another parameter named mode. Then we can pass "NN" / "Bilinear" to it, because almost every framework do it like this. :-)

srkreddy1238:
Yeah, I got the basic logic working from python sample now. But I see it's costly on resources now.

I will check on convolution approach.

Do you think we pass output resolution instead of scale Param to allow asymmetric scaling ?

FrozenGene:
You mean we have different scale factor for height / width? I think you could make scale be tuple accept two values. That is (scaleFactorForHeight, scaleFactorForWidth). You could consider convolution approach, but should also provide the interface for UpSampling so that we can pass the mode "NN" / "Bilinear"

srkreddy1238:
Yes, different scaling to add more flexibility.

Consolidated the resize discussion below.
https://discuss.tvmlang.org/t/scaling-design-for-nn-nearest-neighbour-and-bilinear/187

srkreddy1238:
test_graph.test_print_graph_ir ... ok
test_graph.test_gradient ... ./tests/scripts/task_python_test.sh: line 6: 13532 Segmentation fault      (core dumped) python -m nose -v tests/python/unittest
script returned exit code 255

@tqchen need your help here :)

srkreddy1238:
@tqchen , @Huyuwei  & @FrozenGene 
Added more ops to frontend for Inception V3.
Help me with a final review.

@tqchen  : Please help with the CI issue :)

python/nnvm/frontend/tensorflow.py
```diff
@@ -0,0 +1,651 @@
...
+        self._num_param = 0
+        self._input_node = ''
+
+    def from_tensorflow(self, graph):
+        """Construct nnvm nodes from tensor flow  graph definition - GraphDef.
+
+        Follow the tensorflow graph definition to parse and convert it to NNVM.
+        Some of the assumptions listed below.
+
+            -> First Const node will be comsidered as graph input.
```
FrozenGene: comsidered -> considered

srkreddy1238: Done

srkreddy1238:
@tqchen , @Huyuwei Any other review comments ?

srkreddy1238:
@tqchen and @Huwuwei Can you guys spare some time on this ?

srkreddy1238:
Sure, I will replace real image test case by operator specific cases by random input data.

Convolution possibilities are inherited from tensor flow test cases. I feel good to have more possibilities covered.

What do you suggest ?

srkreddy1238:
@Huyuwei Updated test cases. Pls review.

tqchen:

PR #482, opened 2018-05-14
[WIP]Add SSD Support in NNVM

kevinthesun:
Support SSD in NNVM. Add a quick solution so that major part of network can run on various devices and final NMS part will run on CPU.

### DIFF
--- a/python/nnvm/frontend/mxnet.py
+++ b/python/nnvm/frontend/mxnet.py
@@ -213,6 +213,7 @@ _convert_map = {
     '_plus_scalar'  : _rename('__add_scalar__'),
     '_rdiv_scalar'  : _rename('__rdiv_scalar__'),
     '_rminus_scalar': _rename('__rsub_scalar__'),
+    '_contrib_MultiBoxPrior' : _rename('multibox_prior'),
     'Activation'    : _activations,
     'BatchNorm'     : _batch_norm,
     'BatchNorm_v1'  : _batch_norm,
@@ -237,7 +238,7 @@ _convert_map = {
     'min_axis'      : _rename('min'),
     'reshape'       : _reshape,
     'sum_axis'      : _rename('sum'),
-    'UpSampling'    : _upsampling
+    'UpSampling'    : _upsampling,
 }
 
 def _convert_symbol(op_name, inputs, attrs,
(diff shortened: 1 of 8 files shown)


tqchen:
cc @PariksheetPinjari909 can you help review this?

tqchen:
cc @zhreshold

tqchen:
an example and testcase is needed

kevinthesun:
I'll keep working on this PR and pushing new commits.

python/nnvm/frontend/mxnet.py
```diff
@@ -195,6 +195,22 @@ def _upsampling(inputs, attrs):
     new_attrs = {'scale':int(scale)}
     return _get_nnvm_op('upsampling')(inputs[0], **new_attrs)
 
+def _contrib_MultiBoxDetection(inputs, attrs):
```
PariksheetPinjari909: use snake case for method name `_contrib_multibox_detection`

src/top/nn/object_detection.cc
```diff
@@ -0,0 +1,130 @@
+/*!
+ *  Copyright (c) 2017 by Contributors
+ * \file object_dectection.cc
```
PariksheetPinjari909: Can move to src/top/vision folder

python/nnvm/top/object_detection.py
```diff
@@ -0,0 +1,51 @@
+"""Definition of object detection ops"""
```
PariksheetPinjari909: merge to already existing ` python/nnvm/top/vision.py`

zxwitsme:
Shall we also add MultiBoxTarget op in nnvm even it won't be used in tvm?
Or it will cause error for not implemented op "MultiBoxTarget" when I use the model like this:
load_checkpoint('ssd_mobilenet_512', 0).

liangfu:
@zxwitsme Since MultiBoxTarget is only useful at training stage, it is recommended to export an inference model, instead of using the same model for model training. It can be a lot faster without MultiBoxTarget at inference stage.

zxwitsme:
@liangfu Thanks for the reply. I got the model from here and it includes the MultiBoxTarget op:
https://github.com/zhreshold/mxnet-ssd/releases/mobilenet-ssd-512.zip
How you know how to generate an inference model which doesn't include the training op?

liangfu:
In [this script](https://github.com/zhreshold/mxnet-ssd/blob/master/symbol/symbol_builder.py), there are two models, one for training, the other for inference. Use `get_symbol` when you predict, and save the inference symbol.

zxwitsme:
@liangfu I will try it. Thanks.

tqchen:

PR #487, opened 2018-05-15
[Keras] Cropping2D support

kazum:


python/nnvm/frontend/keras.py
```diff
@@ -289,6 +289,30 @@ def _convert_upsample(insym, keras_layer, _):
...
 
 
+def _convert_cropping(insym, keras_layer, _):
+    _check_data_format(keras_layer)
+    crop_type = type(keras_layer).__name__
+    if crop_type == "Cropping1D":
+        raise NotImplementedError("Cropping1D not implemented")
+    elif crop_type == "Cropping2D":
+        if isinstance(keras_layer.cropping, int):
+            top = bottom = keras_layer.cropping
```
Huyuwei: Better to use crop_t, crop_b, crop_l, crop_r to be consistent with conv and pool

python/nnvm/frontend/keras.py
```diff
@@ -289,6 +289,30 @@ def _convert_upsample(insym, keras_layer, _):
...
+            left = right = keras_layer.cropping
+        else:
+            (_, height, width, _) = keras_layer.input_shape
```
Huyuwei: in_h, in_w

python/nnvm/frontend/keras.py
```diff
@@ -289,6 +289,30 @@ def _convert_upsample(insym, keras_layer, _):
...
+            (height_crop, width_crop) = keras_layer.cropping
```
Huyuwei: crop_h, crop_w

kazum:
@tqchen Okay, I'll start from implementing a slice operator.

python/nnvm/frontend/keras.py
FrozenGene: I am implementing Crop of another FE and find this. Sorry to interrupt you. I just one question: https://github.com/kazum/nnvm/blob/368d69ab4fdf365b9d9298fa23467b226d80abc1/python/nnvm/frontend/keras.py#L302 Should we make it after https://github.com/kazum/nnvm/blob/368d69ab4fdf365b9d9298fa23467b226d80abc1/python/nnvm/frontend/keras.py#L297 immediately? Because it will be used not only 'else' condition but also 'if' condition.

FrozenGene: @kazum @Huyuwei I have implemented Crop in another FE(CoreML) completely. I think I find one bug of our NNVM's transform's bug. When we pass crop's begin be 0, we will get assert error. You can test this case:
x = keras.layers.Cropping2D(cropping=((**0**, 2), (**0**, 4)))(data),  we will get ```src/top/tensor/transform.cc:335: Check failed: param.indices_or_sections[i] > begin (0 vs. 0) indices_or_sections need to be a sorted ascending list``` This is because transformer.cc's ```      CHECK_GT(param.indices_or_sections[i], begin)
          << "indices_or_sections need to be a sorted ascending list";``` and ```    CHECK_LT(begin, dshape[param.axis])
        << "The sum of sections must match the input.shape[axis]";``` we shoule be CHECK_GE / CHECK_LE. Do you want to fix it? I think it is good chance to fix it via Keras. Because we can add related test cases.

kazum: > https://github.com/kazum/nnvm/blob/368d69ab4fdf365b9d9298fa23467b226d80abc1/python/nnvm/frontend/keras.py#L302 Should we make it after https://github.com/kazum/nnvm/blob/368d69ab4fdf365b9d9298fa23467b226d80abc1/python/nnvm/frontend/keras.py#L297 immediately? Because it will be used not only 'else' condition but also 'if' condition.

Okay.

> we shoule be CHECK_GE. Do you want to fix it? I think it is good chance to fix it via Keras.

It looks like a more generic bug and I think it's better to be fixed in another PR.  I'll send the fix later.

tqchen:

PR #491, opened 2018-05-17
don't rename onnx input tensors

souptc:
Currently the onnx frontend will rename the input variables into string like "input_0"/"input_1"/.... 
This is quite confusing, as onnx user knows nothing about how to map an onnx input variable to "input_0"/"input_1"/...., so I think it is better don't do this renaming.


tqchen:
@zhreshold can you review this? do you think if it will affect some of current assumption?

zhreshold:
One concern is how do we get the input name list of a onnx model, I have missed quite a lot onnx changes since long time ago.

souptc:
onnx has a "input" method to get list of input variables names. The only issue is onnx define the "inputs" as the union of graph inputs and weights (I don't like it either, but there is some usage that user want to override weights during inference.), but it should be easy to filter out the weights and return the real graph inputs.

PR #493, opened 2018-05-17
[FRONTEND][Keras] fix reshape

kazum:
This PR fixes the following issues in Reshape.

* The `target_shape` parameter does not include a batch_size dimension.  We have to add it manually.
* Usually, Keras users pass a shape in the channels_last order to the `target_shape` parameter.  It's better to arrange it in the channels_first order, which is used in NNVM.
This fix works only when the above assumption is correct, but will work in more case than the current implementation.
* keras.layers.Reshape doesn't have a `member` attribute in any version of Keras.  Let's simplify the code.


tqchen:
@Huyuwei @PariksheetPinjari909 can you review this?

Huyuwei:
@kazum I am afraid the next layer of reshape will generate error. What kind of network do you use that has a reshape layer? Can you add a test of that kind of network?

kazum:
> In the test case, what if reshape from (32,32,3) to (16,2,16,2,3)?

Then, reshape will change the input shape from (batch, 3, 32, 32) to (batch, 3, 16, 2, 16, 2).

> What kind of network do you use that has a reshape layer?

I'm trying to build MobileNet.  The below code doesn't work without fixing reshape.
```
import keras
import nnvm

keras_model = keras.applications.mobilenet.MobileNet(weights=None)
sym, params = nnvm.frontend.from_keras(keras_model)

shape = (1, 3, 224, 224)
deploy_graph, lib, params = nnvm.compiler.build(sym, target="cuda", shape={"data": shape},
                                                dtype="float32", params=params)
```

PariksheetPinjari909:
I will review this tomorrow.

tqchen:
@Huyuwei any updated comments?

python/nnvm/frontend/keras.py
```diff
@@ -345,11 +346,10 @@ def _convert_concat(insym, keras_layer, _):
...
 def _convert_reshape(insym, keras_layer, _):
-    shape = keras_layer.shape if hasattr(keras_layer, 'shape') \
-       else keras_layer.target_shape if hasattr(keras_layer, 'target_shape') \
-       else None
-    if shape is None:
-        raise TypeError("No shape attribute in reshape layer: {}".format(keras_layer))
+    # XXX: This works only when the target shape is in `channels_last` layout.
+    logging.warning('Assumes that the target shape is in `channels_last` layout.')
+    ch = keras_layer.target_shape[-1]
+    shape = (-1, ch) + keras_layer.target_shape[:-1]
```
Huyuwei: seems here the logic should be:
ch = keras_layer.input.shape[-1].value
assert ch == keras_layer.target_shape[-1], "only supports last dimension in target shape being equal to the channel number of input tensor"
shape = (-1, ch) + keras_layer.target_shape[:-1]

@kazum Can you verify it?

kazum: Thanks, it looks good to me.

Issue #489, opened 2018-05-16
NNVM fails to convert the onnx graph


rajh619:
Loading pretrained onnx model using nnvm.frontend.from _onnx() fails with error.
The model is based on MNIST dataset using chainer ,onnx-chainer framework example

onnx graph representation :

graph Graph (
%1828376209840[FLOAT, 1x784]
) initializers (
%/l3/b[FLOAT, 10]
%/l3/W[FLOAT, 10x100]
%/l2/b[FLOAT, 100]
%/l2/W[FLOAT, 100x100]
%/l1/b[FLOAT, 100]
%/l1/W[FLOAT, 100x784]
) {
%1828376210792 = FC[axis = 1, axis_w = 1](%1828376209840, %/l1/W, %/l1/b)
%1828376210960 = Relu(%1828376210792)
%1828376211072 = FC[axis = 1, axis_w = 1](%1828376210960, %/l2/W, %/l2/b)
%1828376211240 = Relu(%1828376211072)
%1828376211296 = FC[axis = 1, axis_w = 1](%1828376211240, %/l3/W, %/l3/b)
return %1828376211296
}

nnvm.frontend.from_onnx() error :

File “C:\NNVM\NNVM\master\nnvm\python\nnvm_ctypes\symbol.py”, line 181, in creator
ctypes.byref(sym_handle)))
File “C:\NNVM\NNVM\master\nnvm\python\nnvm_base.py”, line 75, in check_call
raise NNVMError(py_str(_LIB.NNGetLastError()))

nnvm._base.NNVMError: Required parameter units of int is not presented, in operator dense(name="")

Looks like dense operator needs  hidden units of the dense transformation ? 

Any help is much appreciated
Thanks

srkreddy1238:
units are inferred inside the front end.

Pls ref https://github.com/dmlc/nnvm/blob/d182c5398096e23346661ae8902a9ecd1d6d238c/python/nnvm/frontend/onnx.py#L385

rajh619:
Hi 

Looks like "  _fully_connected(opset) "   method for dense operator is unused ?
(which internally calls " _infer_channels(... )" )
https://github.com/dmlc/nnvm/blob/d182c5398096e23346661ae8902a9ecd1d6d238c/python/nnvm/frontend/onnx.py#L396

So as temporary fix changed: 
[...]

PR #495, opened 2018-05-17
Include infer_channels utility for 'FC' in convertmap

rajh619:
" _fully_connected(opset) " method for 'dense' operator is unused ,which updates attr['channels'] before sending to AttrConverter .
This solve onnx model load failure as mentioned in issue #489


tqchen:

PR #506, opened 2018-05-22
Squeeze : Infer Shape bug fix (Calculation correction to be inline with compute)

srkreddy1238:
nnvm._base.NNVMError: [13:17:25] src/compiler/compile_engine.cc:210: Check failed: out[i].ndim() == out_info[i].ndim() (2 vs. 4) squeeze


tqchen:
Please add a regression testcase

srkreddy1238:
Test case added.

